Issue #14, opened 2017-12-07
non-remake.yml files


ldecicco-USGS:
Not sure if this is a `scipiper` or `remake` thing, but, I though you mentioned you thought you fixed this. I can't run:
```
scmake(remake_file = "get_data_ready.yml", target_names = "all")
Error: while reading 'remake.yml'
The file 'remake.yml' does not exist
```
Messing around with caching and hoping to separate workflows.

aappling-usgs:
yes, i also thought i fixed that. could you provide a traceback?

ldecicco-USGS:
```
scmake(remake_file = "get_data_ready.yml")
Starting build at 2017-12-07 13:04:19
<  MAKE > all
[    OK ] lib/cfg/gd_config.yml
[ BUILD ] 1_get_data_ready/out                    |  dir.cre...
[ BUILD ] 1_get_data_ready/out/USGS_WQ_DATA_02... |  gd_conf...
Error: while reading 'remake.yml'
The file 'remake.yml' does not exist

> traceback()
31: stop(e)
30: h(simpleError(msg, call))
29: .handleSimpleError(function (e) 
    {
        restore(target$name, path)
        stop(e)
    }, "while reading 'remake.yml'\nThe file 'remake.yml' does not exist", 
        quote(NULL))
28: stop(sprintf("while reading '%s'\n%s", filename, e$message), 
        call. = FALSE)
27: value[[3L]](cond)
26: tryCatchOne(expr, names, parentenv, handlers[[1L]])
25: tryCatchList(expr, classes, parentenv, handlers)
24: tryCatch(yaml_load(read_file(filename)), error = catch_yaml)
23: yaml_read(filename)
22: read_remake_file(obj$file)
21: remake_new(remake_file, verbose, load_sources, allow_missing_packages)
20: remake(remake_file, verbose = FALSE, load_sources = FALSE)
19: remake::list_targets(remake_file = remake_file)
18: indicator_extension()
17: is_indicator(ind_file)
16: as_data_file(ind_file)
15: gd_confirm_posted(ind_file = "1_get_data_ready/out/USGS_WQ_DATA_02-16.xlsx.ind")
14: eval(target$command, envir)
13: eval(target$command, envir)
12: withCallingHandlers(eval(target$command, envir), message = function(e) if (quiet) invokeRestart("muffleMessage"))
11: target_run(target, store, quiet)
10: withCallingHandlers(target_run(target, store, quiet), error = function(e) {
[...]

PR #18, opened 2017-12-08
separate ind_ext from remake file

### LINKED ISSUE #16: move the default indicator extension out of the default remake file
should help resolve #14

aappling-usgs:
resolves #16, which should help with #14 
* there should no longer be an `indicator_extension` target in remake files
* removed `indicator_extension()`, replaced with `ind_ext` arguments in many functions
* default extension is still `ind`; added option to `.Options` to change the default `ind_ext` for a session/project if desired
* renamed for greater consistency: `is_indicator()`->`is_ind_file()`, `as_indicator()`->`as_ind_file()`, `indicator` args -> `ind_file` args
* also renamed for greater consistency: `gd_make_indicator()`->`gd_indicate()` and `sd_make_indicator()`->`sd_indicate()` (like `sc_indicate`)
* added `remake_file` argument to `sc_retrieve`, which was not an indicator thing but was important & missing

hoping that i've caught all the places where the indicator extension is needed or specified, but this was a hairy refactor, so we should be vigilant going forward


ldecicco-USGS:
Hoo-yee....I didn't appreciate how big this change would be!

aappling-usgs:
Yeah, turned out to be a pretty big refactor. Better to do it now before we have a bunch of code that relies on it, though.

R/01-scipiper.R
```diff
@@ -11,9 +11,78 @@
...
+#' \item{`scipiper.ind_ext`}{The indicator file extension to recognize, i.e.,
+#' the final file extension of files for which `is_ind_file()` should return
+#' `TRUE` and which `as_ind_file` should append to the data file name. Used as
+#' the default for the `ind_ext` argument in many functions.`}
+#'
+#' }
+#'
+#' @docType package
+#' @name options
+#' @md
```
ldecicco-USGS: what's this?

R/google_drive.R
```diff
@@ -9,7 +9,7 @@
 #' @param config_file character name of the YAML file where this configuration
 #'   information should be written
 #' @export
-gd_config <- function(folder, config_file=options("scipiper.gd_config_file")[[1]]) {
+gd_config <- function(folder, config_file=getOption("scipiper.gd_config_file")) {
```
ldecicco-USGS: This definitely seems nicer

R/01-scipiper.R
aappling-usgs: https://cran.r-project.org/web/packages/roxygen2/vignettes/markdown.html

aappling-usgs:
Thanks for looking this over, @ldecicco-USGS . Nice to be developing collaboratively on this.

R/01-scipiper.R
ldecicco-USGS: Huh! Never noticed that option. Are you just putting that in by habit, or is there some markdown in there that I'm not noticing? If by habit...the link you just sent me to says you can put this:
```
Roxygen: list(markdown = TRUE)
```
in the DESCRIPTION

Seems like if you are using markdown...you could simplify that item list to be markdowny (not saying you need to....just getting excited about using markdown in my future docs.....)

aappling-usgs: Nice, thanks for pointing that out. I started using `@md` when introducing markdown to more developed packages, but scipiper seems like a good place to use the package-wide option. Switching over right now...

aappling-usgs: Switched the `@md` tags. Just noticed that I could probably switch the itemized list, too, but i'm gonna pass on that for now because I should be finishing other tasks today. Another time

PR #19, opened 2017-12-08
I need this...

ldecicco-USGS:


ldecicco-USGS:
It was returning the data path if it was only up on GD and not if it was there locally

aappling-usgs:
Oops! good catch.

PR #24, opened 2018-01-25
bringing s3 functions in line with google drive

wdwatkins:


R/s3.R
```diff
@@ -29,95 +31,135 @@ s3_config <- function(path, bucket, profile='default', config_file=getOption("sc
...
+  bucket_contents <- aws.s3::get_bucket_df(bucket = s3_config$bucket)
+  exists_on_s3 <- local_file %in% bucket_contents$Key
+  
+  #upload to S3 - note that S3 is a flat file system, so folders don't need
+  #to be created.  'Directories' are just part of object keys
+  match.arg(on_exists)
+  if(exists_on_s3 && on_exists == "stop") {
+    stop('File already exists and on_exists==stop')
+  } else {
+    if(verbose) message("Uploading ", local_file, " to S3")
```
aappling-usgs: even though the `aws.s3::put_object` call is the same for both a fresh upload and a replacement (at least, i assume so based on your code - is that right?), it would be nice to have the message distinguish between those two options. could you have this message choose between "Uploading" and "Replacing" to indicate which one is happening?

R/s3.R
```diff
@@ -133,45 +175,16 @@ s3_confirm_posted <- function(
...
-s3_indicate <- function(ind_file, remote_time) {
-  
-  warning('s3_indicate is out of date relative to gd_indicate') # need to update s3_xx with lessons learned from gd_xx
-  
-  # write the cache file
-  if(is.character(remote_time)) remote_time <- s3_read_time(remote_time)
-  writeLines(remote_time, con=ind_file)
-  
+  bucket_contents <- aws.s3::get_bucket_df(bucket = s3_config$bucket)
+  remote.info <- filter(bucket_contents, Key == data_file)
```
aappling-usgs: let's stick to underscore_separation for variable names

R/google_drive.R
```diff
@@ -363,38 +356,21 @@ gd_confirm_posted <- function(
...
-#' @param config_file character name of the YAML file containing project-specific
-#'   configuration information
+#' @param md5_checksum character md5 hash
 #' @keywords internal
-gd_indicate <- function(ind_file, md5_checksum) {
-  
-  # we could prepare a timestamp...but checksum is even better
-  # if(is.character(remote_time)) remote_time <- POSIX2char(gd_read_time(remote_time))
-  
+indicate <- function(ind_file, md5_checksum) {
```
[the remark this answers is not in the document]
wdwatkins: Yeah, `indicate` is basically empty now, so it should go.  Don't see any downsides.

R/google_drive.R
```diff
@@ -356,21 +356,7 @@ gd_confirm_posted <- function(
   remote_info <- remote_path %>% slice(n()) %>% pull(drive_resource) %>% .[[1]]
   
   # we could prepare a timestamp from modifiedTime...but checksum is even better
-  indicate(ind_file, md5_checksum=remote_info$md5Checksum)
+  sc_indicate(ind_file, md5_checksum=remote_info$md5Checksum)
```
aappling-usgs: current behavior is to use the key `hash` instead of `md5_checksum` in sc indicator files - i think all that's needed is to change the argument name

R/s3.R
```diff
@@ -179,11 +179,11 @@ s3_confirm_posted <- function(
...
   s3_config <- yaml::yaml.load_file(config_file)
   bucket_contents <- aws.s3::get_bucket_df(bucket = s3_config$bucket)
-  remote.info <- filter(bucket_contents, Key == data_file)
-  if(nrow(remote.info) == 0) {
+  remote_info <- filter(bucket_contents, Key == data_file)
+  if(nrow(remote_info) == 0) {
     stop(paste0("failed to find S3 file with Key=", data_file))
   } else {
-    indicate(ind_file, md5_checksum = remote.info$ETag)
+    sc_indicate(ind_file, md5_checksum = remote_info$ETag)
```
aappling-usgs: again `hash` instead of `md5_checksum`

aappling-usgs:
Cool, I think that's everything, right?

wdwatkins:
should be :shipit:

Issue #23, opened 2018-01-24
make s3_put and gd_put smarter?


wdwatkins:
S3_put and gd_put could check the file hashes before sending files over the wire.  This may be redundant with what remake already knows from indicator files though.

aappling-usgs:
we've been discussing on slack. i'm convinced that this would be an improvement and that our main reason for not doing it right away is that other things have higher priority, and that the improvement is only one of efficiency rather than a change in how pipelines work. to make the case for implementing this feature, if/when we do it:

This would be useful to the 2-target pattern, where we have one target to create and post a data file and a second target to download the file from the remote cache. e.g.,
```
## remake file
# create and post 1_data/out/mydata.rds (and an indicator for it) at once
1_data/out/mydata.rds.ind:
  command: create_and_post_mydata(target_name)
# retrieve data file on demand
1_data/out/mydata.rds:
  command: gd_get('1_data/out/mydata.rds.ind')

## function definitions
create_and_post_mydata <- function(ind_file) {
  # create data file (no need to make indicator yet)
  data_file <- as_data_file(ind_file)
  mydata # <- ...compute mydata here...
  write.csv(mydata, data_file)
  # post and create indicator file
 gd_put(remote_ind=ind_file, local_source=data_file)
}
```
in the above example, there could be cases where the inputs to `create_and_post_mydata` have changed but the data file remains the same. in those situations, it would be ideal if the call to `gd_put` recognized the redundancy and didn't bother posting the same file again. (same for `s3_put` in a parallel scenario.)

The reason I'm fine with waiting on this issue is that for 3-target patterns, remake already provides that same efficiency. example:
```
## remake file
# create 1_data/cache/mydata.rds (and an indicator for it) locally
1_data/tmp/mydata.rds.ind:
  command: create_mydata(target_name)
# post 1_data/cache/mydata.rds to 1_data/out/mydata.rds on Drive
1_data/out/mydata.rds.ind:
  command: gd_put(remote_ind=target_name, local_source='1_data/tmp/mydata.rds.ind')
# retrieve data file on demand
1_data/out/mydata.rds:
  command: gd_get('1_data/out/mydata.rds.ind')

## function definitions
create_mydata <- function(ind_file) {
  data_file <- as_data_file(ind_file)
  mydata # <- ...compute mydata here...
  write.csv(mydata, data_file)
  sc_indicate(ind_file, data_file=data_file)
}
```
[...]

PR #28, opened 2018-02-07
retry gd_confirm_posted when needed, reduce unnecessary calls to Drive

aappling-usgs:
### The problem Matt was having today:
In a bunch of calls to `gd_put` in a scipiper task loop, occasionally some of the `gd_put` calls were failing with a message such as,
```
Error in gd_confirm_posted(ind_file = remote_ind, config_file = config_file) : 
  File was not found on Google Drive: 1_wqdata/out/wqp/Virgin_Islands_tss_001.feather
```
where the file `1_wqdata/out/wqp/Virgin_Islands_tss_001.feather` was actually successfully posted to Drive (we can see it there), but somehow `gd_confirm_posted` (which gets called within `gd_put` after the posting) wasn't finding the file. We hypothesize that this occurred simply because sometimes it takes Drive a few seconds to notice that it has the new file and to make the file name available to a query (`drive_ls` in `gd_confirm_posted`).

### SO, this PR has these fixes (we hope) to the problem:
* create an internal function, `retry_patiently`, to try and retry evaluating an expression with progressively longer waits between tries
* use `retry_patiently` in the call to `gd_confirm_posted` in `gd_put`, so that if an immediate query doesn't work, maybe sleeping and retrying will let us find the file. The current wait times between tries are 0  1  3  9 27 81 seconds.
* implement #23 on the google drive side (still need an S3 version of this) - i.e., within gd_put, don't repost a file to Drive unless the existing file on Drive is genuinely different, according to its md5 checksum. This is helpful because Matt and I have already posted a bunch of files, and the first two changes above will cause remake to call gd_put on all of those files just because gd_put has changed, and we don't actually want to re-push a bunch of identical files to Drive.
* as a second piece of #23 for Drive, don't call `gd_confirm_posted` in the special case where we've just located an existing file on Drive and confirmed that it is identical to the local file; in this case, the remote_hash we just retrieved is already the most up-to-date info, so we can avoid a second call to Drive.

### Smaller things I changed:
* make note in `?gd_put` about what we've learned about personal vs team google drive & 'replace' - in short, 'replace' is a bad option if multiple people are going to be posting/replacing files on Drive, at least if you're using a personal rather than Team drive.
* get rid of those annoying "102 items so far" messages in calls to `drive_ls` within `gd_locate_file`. just too many of them in a remake context.
* `scdel` had `verbose` as the 2nd argument, `remake_file` as the third. this is inconvenient if you want to force-rebuild a target by alternating `scmake` and `scdel`, e.g., `scdel('1_wqdata/out/wqp/Virgin_Islands_tss_001.feather.ind', 'tasks_1_wqp.yml')` and then `scmake('1_wqdata/out/wqp/Virgin_Islands_tss_001.feather.ind', 'tasks_1_wqp.yml')`. I don't think many/any projects have calls to `scdel` built in, so I'm making the switch before it's too late.

### Example:
When the local file is different from the remote file, and if verbose=TRUE in call to gd_put, then you should see:
```
[ BUILD ] 1_wqdata/out/wqp/Virgin_Islands_tss_001.feather.ind             |  gd_put(...
Updating 1_wqdata/tmp/wqp/Virgin_Islands_tss_001.feather on Google Drive
Trying "gd_confirm_posted(ind_file = remote_ind, config_file = config_file)":
  Waiting 0.0 seconds before evaluating
Finished build at 2018-02-07 16:14:22
Build completed in 0.35 minutes
```
(and Virgin_Islands_tss_001.feather.ind gets written or rewritten locally with the hash of the newly posted file).

When the local and remote files are identical and verbose=TRUE, then:
```
[ BUILD ] 1_wqdata/out/wqp/Virgin_Islands_tss_001.feather.ind             |  gd_put(...
Not re-posting identical 1_wqdata/tmp/wqp/Virgin_Islands_tss_001.feather to Google Drive
Finished build at 2018-02-07 16:14:22
Build completed in 0.07 minutes
```
(and Virgin_Islands_tss_001.feather.ind gets written or rewritten locally but with the same contents as before)


R/google_drive.R
```diff
@@ -166,30 +171,50 @@ gd_put <- function(
...
-  gd_confirm_posted(ind_file=remote_ind, config_file=config_file)
+  # write the indicator file
+  if(remote_file_changed) {
+    # most common case. involves another check on Google Drive, with patience in
+    # case Drive doesn't fully recognized the new file right away
+    retry_patiently(gd_confirm_posted(ind_file=remote_ind, config_file=config_file), verbose=verbose)
+  } else {
+    # special case. if the remote file was already identical to the local, then
+    # we did not repost the file, and we don't need to check Drive again
+    sc_indicate(ind_file=remote_ind, hash=remote_hash)
```
wdwatkins: If the remote file is already identical, theoretically the indicator file would also be the same right?

wdwatkins: I suppose rewriting it wouldn't hurt anything though

R/google_drive.R
```diff
@@ -166,30 +171,50 @@ gd_put <- function(
...
-      replace={
-        if(verbose) message("Replacing ", local_file, " on Google Drive")
-        googledrive::drive_rm(googledrive::as_id(remote_id), verbose=verbose)
-        remote_id <- googledrive::drive_upload(media=local_file, path=googledrive::as_id(parent), type=type, verbose=verbose)$id
-      },
-      stop={
-        stop('File already exists and on_exists==stop')
+    if(on_exists == 'stop') {
+      stop('File already exists and on_exists==stop')
+      # remote_file_changed <- FALSE # true but unnecessary
```
wdwatkins: did you intend to leave this?

R/google_drive.R
aappling-usgs: Yeah...I thought about an additional check so that this was only re-written if the local indicator file didn't exist (which could happen if the analyst manually deleted that file, which he/she might do by accident or if he/she wanted to force this function to run again). However, I also think remake might complain if the target goes untouched...wasn't totally sure about that, but seemed safer to assume that remake _would_ complain and just avoid the issue.

R/google_drive.R
aappling-usgs: well...yes, just for clarity about the meaning of `remote_file_changed`, but you're right to imply that it's sorta silly to have it in there. i'll take it out.

PR #29, opened 2018-02-08
point to the existing data_file in gd_put hashing

aappling-usgs:
Whoops! Pointed to the wrong file. Thanks to Matt for discovering the issue.


aappling-usgs:
@wdwatkins it's just a one-liner, but could you review and merge anyway for form's sake?

Issue #54, opened 2018-07-11
gd files do not support `[` in file names


jread-usgs:
Trying to post a file that looks like `5_vizprep/out/storm_point_[20170826-11]_fun.rds`, which fails with
```r
Error in grep(pattern, res_tbl$name) : 
  invalid regular expression '^5_vizprep$|^out$|^storm_point_[20170826-11]_fun\.rds$', reason 'Invalid character range'
```
This happens in `gd_locate_file` with this block

```r
  relevant_files <- bind_rows(googledrive::drive_get(id = googledrive::as_id(gd_config$folder)), 
    googledrive::drive_ls(path = googledrive::as_id(gd_config$folder), 
      pattern = sprintf("^%s$", gsub(".", "\\.", fixed = TRUE, 
        x = gsub("/", "$|^", relative_path))), verbose = FALSE, 
      recursive = TRUE)) %>% dplyr::mutate(parents = lapply(drive_resource, 
    function(dr) {
      parent <- unlist(dr$parents)
      if (is.character(parent)) 
        parent
      else NA
    })) %>% tidyr::unnest(parents, .preserve = c("drive_resource"))
```

wdwatkins:
We just need to escape the brackets, they are being interpreted as a numeric range, e.g. https://www.regular-expressions.info/numericranges.html

jread-usgs:
Can you take this on @wdwatkins ?

wdwatkins:
yeah should be a simple fix

PR #55, opened 2018-07-11
allow brackets in file names on drive

wdwatkins:
The regexing was breaking since the brackets weren't escaped


wdwatkins:
fixes #54

Issue #62, opened 2018-07-24
option to build target[s] without pushing to shared cache


aappling-usgs:
This idea came out of the vizstorm-GIF sprint retro. Because 5 of us were coding simultaneously, we had lots of temporary inconsistencies between the shared cache and the .ind files in the canonical GitHub repo. This happened when someone changed the cache on Drive by building a target locally (one that calls gd_put), but their PR didn't get created or merged for a while after that.

A scipiper feature that could ease this pain would be an option that disables the pushing action in `gd_put` (and `s3_put`) without an explicit change to the function that is calling `gd_put`. It might be nice to have a session option so that you can go into a dev mode where every call to gd_put just passes through without actually posting to Drive, unless you explicitly set `post=TRUE` or the like in your call to `scmake()`, or maybe you just need to explicitly turn dev mode off while you make that final call to scmake() just before your PR is merged.

This was Lindsay's idea. It's a great one.

aappling-usgs:
dry_put is implemented and seems to be working

PR #63, opened 2018-07-27
introduce dry_put option. needs testing

aappling-usgs:
Basic idea is that this PR adds a session option called `dry_put` that, if set to TRUE, allows gd_put and s3_put calls to proceed locally as if everything is fine, but neither call actually pushes anything to gd or s3. This should be useful for projects where everybody is working simultaneously on a repo and we want to keep a tighter synchronization between our shared code base and our shared Drive/S3 cache.

This PR might do the trick for https://github.com/USGS-R/scipiper/issues/62, but I think we need to test it on some real projects before we can call that issue resolved. I would also appreciate some help thinking through the implications - does this theory seem sound? What happens if people alternate between dry_put=TRUE and dry_put=FALSE, especially if they don't explicitly force a remake of the objects that were dry-put? What if they run with dry_put=TRUE and then forget to rerun with FALSE before creating a PR?

### DIFF
--- a/R/s3.R
+++ b/R/s3.R
@@ -91,7 +91,7 @@ s3_put <- function(
   if(isTRUE(dry_put)) {
     sc_indicate(ind_file=remote_ind, warning="dry_put=TRUE; not actually pushed", data_file=data_file)
     mock_move_copy(mock_get, local_file, data_file)
-    return(NA)
+    return()
   }
   
   # prepare to use S3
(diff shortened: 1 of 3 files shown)


R/s3.R
```diff
@@ -84,6 +86,14 @@ s3_put <- function(remote_ind, local_source,  mock_get=c('copy','move','none'),
   # identify the remote data file to be indicated by remote_ind
   data_file <- as_data_file(remote_ind, ind_ext=ind_ext)
   
+  # allow for dry runs of s3_put, where we move and create files locally but do
+  # nothing on S3
+  if(isTRUE(dry_put)) {
+    sc_indicate(ind_file=remote_ind, warning="dry_put=TRUE; not actually pushed", data_file=data_file)
+    mock_move_copy(mock_get, local_file, data_file)
+    return(NA)
```
jread-usgs: Looks like `s3_put` returns an `NA` and `gd_put()` returns `NULL`. Should they both be the same?

PR #67, opened 2018-08-02
1-argument option for gd_put, s3_put

aappling-usgs:
Really tiny PR because I'm sick of having to include two arguments to gd_put or s3_put when in fact one is often plenty. Now you can pass in just one argument, and that argument (`remote_ind`) should be the .ind file, and the corresponding data file you want to push must have exactly the path and data_file name implied by that .ind file.

```r
gd_put('1_fetch/out/mydata.rds.ind')
s3_put('1_fetch/out/mydata.rds.ind')
```

(This is only appropriate if you're both creating and pushing a data file from within the same function, which means a 2-target remake pattern for this shared-cache data file, with one target to create locally + push, one target to retrieve. If you're using a 3-target remake pattern with one target to create locally, one to push, and one to retrieve, you must and can still specify the 2nd argument to gd_put/s3_put so that it points to a data file other than the one indicated by the remote_ind; usually this other data file will be in a /tmp/ directory whereas the remote_ind will point to a similarly named file in the /out/ directory.)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: scipiper
 Title: Support functions for ushering data through a scientific workflow
-Version: 0.0.10
-Date: 2018-07-27
+Version: 0.0.11
+Date: 2018-08-02
 Authors@R: c(person("Alison", "Appling", email = "aappling@usgs.gov", role = c("aut", "cre")),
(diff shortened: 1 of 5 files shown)


jread-usgs:
I'm a fan :+1:

PR #68, opened 2018-08-02
resolve #32 vectorize as_data_file

### LINKED ISSUE #32: vectorize as_data_file
```
as_data_file(c('xx.str.ind','yy.txt.ind'))
[1] "xx.str" "yy.txt"
Warning message:
In if (!is_ind_file(ind_file, ind_ext = ind_ext)) { :
  the condition has length > 1 and only the first element will be used
```
we can do better than this

aappling-usgs:
Another teeny PR - just vectorizing as_data_file and making as_ind_file follow the same warnings format

### DIFF
--- a/R/scmake.R
+++ b/R/scmake.R
@@ -252,9 +252,9 @@ is_ind_file <- function(target_names, ind_ext=getOption("scipiper.ind_ext")) {
 #' as_ind_file('mydata.rds', ind_ext='st') # 'mydata.rds.st'
 #' as_ind_file('mydata.rds.ind') # Error: "data_file is an indicator file already"
 as_ind_file <- function(data_file, ind_ext=getOption("scipiper.ind_ext")) {
-  are_ind_files <- is_ind_file(data_file, ind_ext=ind_ext)
-  if(any(are_ind_files)) {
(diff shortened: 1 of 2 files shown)


R/scmake.R
```diff
@@ -252,9 +252,9 @@ is_ind_file <- function(target_names, ind_ext=getOption("scipiper.ind_ext")) {
 #' as_ind_file('mydata.rds', ind_ext='st') # 'mydata.rds.st'
 #' as_ind_file('mydata.rds.ind') # Error: "data_file is an indicator file already"
 as_ind_file <- function(data_file, ind_ext=getOption("scipiper.ind_ext")) {
```
wdwatkins: is there a convention on whether to pluralize names of function arguments if they are vectors?

aappling-usgs: Hmm, the usually-vectorized nature of R blurs that line a lot. Here are the first few examples I can find with named, vectorized arguments:
* `dir(path, ...)`, where path can be a vector
* `log(x)` and `exp(x)`, where `x` can be a vector
* `ifelse(test, yes, no)` where all 3 args can be vectors

So I think the answer is no, or if there is such a convention, it's not widely followed?

Issue #65, opened 2018-08-01
improve efficiency of loop_tasks is_current check


aappling-usgs:
I was just looking at https://github.com/USGS-R/scipiper/blob/master/R/loop_tasks.R#L70-L76 again and am not sure it's as efficient as I initially thought. In particular, we call `get_remake_status(targets, task_makefile)` on _all_ possible targets and only later exclude those that already have an existing file. But the check for whether existing files are current is the slow part, so we surely shouldn't be calling `get_remake_status(targets, task_makefile)` on those existing files at all, right?

And maybe we should make it a user option to either loop over all tasks or to loop over complete unbuilt tasks (for which file doesn't yet exist)? Sometimes the hashing is expensive while the retries of unbuilt files are the focus; other times the hashing is cheap and we'd like to include retries for built but out-of-date files (e.g., when we forget or don't know to or don't want to delete the files before building).

jread-usgs:
Yes, I think I ran into this taking a long time to run when I had a lot of targets. I wasn't thinking much about the inefficiencies in `which_incomplete` as much as us needing to call that again with each `try` in the loop (I am dealing w/ a failure-prone build). Since we are catching errors within the try, we know which targets succeed and which fail, so can `incomplete_targets <- which_incomplete(targets, task_makefile)` be set once before the first loop and then updated within the `try`s after each `scmake` call?

aappling-usgs:
Ooh, that's also a good idea!

jread-usgs:
I've got this (call `which_incomplete` only once) in the work I am doing for `progress` bar if you want to take a look at it w/ that PR. 

I still think there are gains to be made in the internals of `which_incomplete`

PR #69, opened 2018-08-03
Task message

jread-usgs:
This PR has a touch of #65 in it, but is mostly designed to provide an option of much more compact output of `loop_tasks`, which I find to be a really useful function when I am doing a bunch of failure-prone tasks (like fetching data from brittle services/servers). The utility of this is only apparent when you have hundreds/thousands of targets and a decent likelihood of failure. I added the "silent" option, which I'm not sure we should include...since the final `scmake` that is called to verify completeness is not silent (is still `verbose`), but thought it would be a good idea to implement it to have the complete cases covered (verbose, ~compact, and silent). If we want to go all out, we could use that "silent" input to set `verbose` in the final `scmake()` call to `FALSE`. 

But I wanted to get feedback on this PR before messing around with that. 

using `task_msg = 'progress'`:
task_progress

using `task_msg = 'verbose'` (the default, and this should not have changed):
task_verbose

### DIFF
--- a/R/loop_tasks.R
+++ b/R/loop_tasks.R
@@ -31,9 +31,9 @@
 #'   failed task. Especially useful if the error was likely to be inconsistent
 #'   (e.g., a temporary network issue) and might not occur again if we wait a
 #'   while
-#' @param task_mssg "vebose","progress", or "silent" to define the format of task 
+#' @param task_msg "vebose","progress", or "silent" to define the format of task 
 #'   messages. Use "verbose" for a return-line `message` for each task, "progress" 
-#'   for a compact progress bar for task status, and "silent" for no output. 
+#'   for a compact progress bar for the status of each task, and "silent" for no output. 
 #'   Using "silent" is not recommended, but is implemented here for completeness. 
 #' @export
 #' @import progress
@@ -42,7 +42,7 @@ loop_tasks <- function(
   task_names=NULL, step_names=NULL,
   num_tries=30, sleep_on_error=0,
   ind_ext = getOption('scipiper.ind_ext'),
-  task_mssg = c('verbose','progress','silent')) {
+  task_msg = c('verbose','progress','silent')) {
   
   # provide defaults for task_names (all tasks) and step_names (final_steps)
   target_default <- yaml::yaml.load_file(task_makefile)$target_default
@@ -58,7 +58,7 @@ loop_tasks <- function(
     step_names <- attr(task_plan, "final_steps")
   }
   
-  task_mssg <- match.arg(task_mssg)
+  task_msg <- match.arg(task_msg)
   # identify the task-step targets to be run, ordered by tasks and then steps
   # within tasks
   targets <- unlist(lapply(unname(task_plan[task_names]), function(task) {
@@ -82,6 +82,13 @@ loop_tasks <- function(
     incomplete_targets <- which(!(file.exists(targets) | is_current))
   }
   
+  incomplete_targets <- which_incomplete(targets, task_makefile)
+  if (task_msg == 'progress'){
+    pb <- progress::progress_bar$new(
+      format = "  Built :what [:bar] :percent",
+      clear = FALSE, total = length(incomplete_targets))
+  }
+  
   # run the targets in a loop, with retries, so that we complete (or skip) one
   # task before trying the next. If we just ran scmake(job_target,
   # task_makefile) directly, remake would try to build the first step for all
@@ -95,27 +102,27 @@ loop_tasks <- function(
     # will catch it in the final calls of this function when we run remake on
     # the entire job, properly. if a target is an object rather than a file,
     # we'll build it every time
-    incomplete_targets <- which_incomplete(targets, task_makefile)
     num_targets <- length(incomplete_targets)
+    target_complete <- rep(FALSE, num_targets)
+    
     if(num_targets == 0) break
     
     # if there are remaining targets, try to run them
-    message(sprintf(
-      "\n### Starting loop attempt %s of %s with %s tasks remaining:",
-      this_try, num_tries, num_targets))
-    this_try <- this_try + 1
-    
-    if (task_mssg == 'progress'){
-      pb <- progress::progress_bar$new(
-        format = "  Building :what [:bar] :percent",
-        clear = TRUE, total = num_targets)
+    loop_start_msg <- sprintf(
+      "### Starting loop attempt %s of %s with %s tasks remaining:",
+      this_try, num_tries, num_targets)
+    if (task_msg == 'verbose'){
+      message(paste0("\n", loop_start_msg))
+    } else if (task_msg == 'progress'){
+      pb$message(loop_start_msg)
     }
+    this_try <- this_try + 1
     
     for(i in seq_len(num_targets)) {
       tryCatch({
         target_num_overall <- incomplete_targets[i]
         target <- targets[target_num_overall]
-        if (task_mssg == 'verbose'){
+        if (task_msg == 'verbose'){
           message(sprintf(
             "Building task %s of %s in loop: %s (#%s of %s total)",
             i, num_targets, target, target_num_overall, num_targets_overall))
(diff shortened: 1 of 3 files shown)


aappling-usgs:
Cool!

Some quick early questions (followed by suggestions):
* Why does it try `#3 of 6` after having already tried `#4` and `#5` in loop 1?
image
* Why is `#6` being [re]tried in loop 2 when there's no failure record for `#6` in loop 1?
* When it's done with `#6` in loop 2, what does it mean that the progress bar is at 67%? (surely `#6` is the 2nd of the 2 remaining tasks in that loop?)
image

Suggestions:

1. I'm not crazy about how the task numbers are showing up in the messages within the individual loops. I think some of my questions above reflect this, and I am also confused by `#6 of 6` with `2 tasks remaining`. I think the `total` specification helps in the verbose option, though I'm also happy for you to revise the original messaging scheme (I like the compression of the error message to 1 line, too). Anyway, instead of numbers within the loops, could we use the task-step names defined in the task plan? This also raises the opportunity to add something I've been wanting to add, which is a printout of exactly which `scmake()` command you should call if you want to directly retry just one failed target. Are you up for adding that? The resulting error message could be something like:
```r
Error in prep("IL"): Data fetch failed for IL; debug with scmake("states/log/IL.ind", "6_drivers_fetch_tasks.yml")
```

2. Could the progress bar show which task is being attempted rather than which one was most recently completed? Like this:
```
  Building states/log/DC.ind (#2 of 6): [====--------------------] 17%
```

3. The combination of 2 tasks remaining and a progress bar at 5/6 is unintuitive to me, though on my 25th view of the GIF I'm starting to at least understand it. Would you be up for having the progress bar reflect progress through the current loop rather than through the total number of tasks remaining as of the start of the loop_tasks call? Then I think the current messaging format could be similar but would show
```
  Building states/log/PA.ind (#2 of 2): [============------------] 50%
```
or
```
  Building states/log/PA.ind (#2 of 2 in loop 2): [============------------] 50%
```
or even 2 simultaneous progress bars if it's important to show both current and total progress?
```
  Building states/log/PA.ind (#2 of 2 in loop 2): [============------------] 50%
  Overall task plan progress (22 of 24 total): [======================--] 92%
```


4. I agree that the `silent` option should set `verbose=FALSE` anytime `scmake()` is called.

R/loop_tasks.R
```diff
@@ -88,29 +102,50 @@ loop_tasks <- function(
...
+          pb$tick(tokens = list(
+            what = sprintf('%s (#%s of %s)', 
+                           target, target_num_overall, num_targets_overall)))
+        }
       }, error=function(e) {
-        message(sprintf("  Error in %s: %s", deparse(e$call), e$message))
-        message(sprintf("  Skipping task #%s due to error", incomplete_targets[i]))
+        if (task_msg == 'verbose'){
+          message(sprintf("  Error in %s: %s", deparse(e$call), e$message))
+          message(sprintf("  Skipping task #%s due to error", incomplete_targets[i]))
```
aappling-usgs: fine with me if you want to combine these into 1 line. might look funny when the error is long, but probably worth it

aappling-usgs: And then we could have the same message regardless of the value of task_msg, keep the code simple

jread-usgs:
Responses:
> Why does it try #3 of 6 after having already tried #4 and #5 in loop 1?
That is another example of below, the past tense of "built". It isn't building 3, 3 was just the last successful target build at that time until it builds 6
> Why is #6 being [re]tried in loop 2 when there's no failure record for #6 in loop 1?
it isn't being retried, the message that "#6" was built is being carried over into the start of loop 2. I hadn't noticed this, but I can see it being confusing. These messaged differ from the verbose option between "building" and "built"...so they are telling us slightly different things
> When it's done with #6 in loop 2, what does it mean that the progress bar is at 67%? (surely #6 is the 2nd of the 2 remaining tasks in that loop?)
this is also the same issue w/ the last successful build showing up in the progress bar message. targets 1,2,3 &6 were built in loop one, so starting loop 2, it is 67% complete with the overall set of targets

> I'm not crazy about how the task numbers are showing up in the messages within the individual loops. I think some of my questions above reflect this, and I am also confused by #6 of 6 with 2 tasks remaining. I think the total specification helps in the verbose option, though I'm also happy for you to revise the original messaging scheme (I like the compression of the error message to 1 line, too)
Agreed. I thought I was following the naming convention that existed in `verbose` other than removing "total". I found the verbose messages confusing the first few times around the block w/ `loop_tasks`. I think it may be easier to understand if these messages are the task-steps, but some of my task steps are really long and would require truncating (and that is hard to do automatically w/o missing the key characters each step has that makes it unique). 

> This also raises the opportunity to add something I've been wanting to add, which is a printout of exactly which scmake() command you should call if you want to directly retry just one failed target. Are you up for adding that? The resulting error message could be something like:
seems like a good idea

> The combination of 2 tasks remaining and a progress bar at 5/6 is unintuitive to me, though on my 25th view of the GIF I'm starting to at least understand it. 
hmm...are you talking about this:
image

When the message to start loop two shows up, there are two tasks remaining and it is at 4/6, then a target is built and it bumps to 5/6. Is the message above that remains confusing?

For comments about the progress within the loop vs overall, I think the overall progress bar is the one that I find most useful. 🤔 about two bars..

aappling-usgs:
I'm OK with just an overall progress bar as long as it says "total" or "overall" somewhere on that line. Maybe it would also help to have the loop-starting messages clarify "remaining" relative to the total? 
```
### Starting loop attempt 2 of 30 with 2 of 6 tasks remaining:
```

Is "building" an option with the progress bar? I really think I'd find that easier to understand than "built". From your responses, looks like most of my confusion stemmed from that one word.

What about if we used the conceptual task and/or step names rather than the target names in the status messages? (I'm imagining that target names are what you're thinking of when you say "some of my task steps are really long") That might help with long task names. The current message already shows the [potentially long] target name in the `Built [targetname] (#a of b) [====---]` lines, so we could back off to `Building [taskname]` or `Building [taskname], [stepname]` rather than using the target names?

I do still think it'd be worth spelling out the target name in the error messages `debug with...` segment, if you're up for that. Hopefully error messages will usually be infrequent enough that this is more valuable than obnoxious - I'm quite sure that's the case for some task tables. And/or we could add a 4th level of verbosity to distinguish between compact and compact-with-debug-command?

aappling-usgs:
Also, does `progress` allow you to put the current task-step info on a separate line from the progress bar? Then it'd still be taking up just 2 lines rather than expanding number, but you wouldn't have to worry so much about truncation.

jread-usgs:
> Also, does progress allow you to put the current task-step info on a separate line from the progress bar? 

i couldn't find a clear way to do this. 

> Is "building" an option with the progress bar? 

doing this, yes.

>  "some of my task steps are really long"

yes, this would help for me since it would get rid of the directory, but my tasknames are pretty long:
```
# --- NLDAS_time[000000.024999]_x[221.344]_y[132.196]_var[apcpsfc].nc --- #
  
  6_drivers/out/nc/NLDAS_time[000000.024999]_x[221.344]_y[132.196]_var[apcpsfc].nc:
    command: nccopy_split_combine(target_name, max_steps = I(100))

  6_drivers/log/nc/NLDAS_time[000000.024999]_x[221.344]_y[132.196]_var[apcpsfc].nc.ind:
    command: sc_indicate(target_name)
    depends:
      - 6_drivers/out/nc/NLDAS_time[000000.024999]_x[221.344]_y[132.196]_var[apcpsfc].nc
```

aappling-usgs:
> my tasknames are pretty long
So they are. I don't know what a good answer is, then. Maybe just have a progress bar with absolutely no annotations when in "compact" mode? Still [provide an option to] print out the scmake() code (with no task number) but only if a task fails, such that a largely successful run would show very little text at all?

R/loop_tasks.R
```diff
@@ -197,6 +196,6 @@ loop_tasks <- function(
       strwrap("Set task_names=NULL, step_names=NULL to build the job target. Until then, expect this error:"),
```
aappling-usgs: Line 190 above could also use an `if(verbose)`

PR #72, opened 2018-08-08
Quicker loop transition

aappling-usgs:
* Deleted a few lines to speed up the time from the main task loop to the final check in loop_tasks
* Adjusted the formatting of the progress bar so it's easier to view messages printed by task-step functions
* Added tests for loop_tasks


aappling-usgs:
Had a heck of a time getting those tests in place - they still don't run successfully with rstudio Test or Check commands, but they do run manually.

R/loop_tasks.R
```diff
@@ -83,7 +83,8 @@ loop_tasks <- function(
   total_num_tasks_complete <- 0
   if (verbose){
     pb <- progress::progress_bar$new(
-      format = ":what [:bar] :percent",
+      format = ":what [:bar] :percent  ",
+      width = getOption('width')+4,
```
aappling-usgs: progress bars don't generally play well with other `message()`s, but this formatting makes it more likely that if you do have other `message` calls within your task steps, they'll at least be readable if not beautifully placed.

R/loop_tasks.R
```diff
@@ -83,7 +83,8 @@ loop_tasks <- function(
...
       show_after = 0, 
       clear = FALSE, total = total_num_tasks_incomplete)
```
aappling-usgs: I haven't changed it, but I'm wondering if `total_num_tasks_incomplete` is the best choice for the progress bar total. What if it were `total_tasks = length(targets)` instead? If you're starting from a completely unbuilt remake task file, the numbers are the same, but suppose you just build 200 of 300 tasks successfully yesterday, before your computer crashed. When you restart today, would it make more sense for you to see 201/300, 202/300, etc. in the progress bar, or 1/100, 2/100, etc? The current implementation is the latter. I have a slight inclination toward the former, but I also suspect there's no universally correct answer here. What are your thoughts, @jread-usgs ?

inst/extdata/sharedcache/demo.R
```diff
@@ -1,116 +0,0 @@
...
-
-make_cache_indicator <- function(indfile) {
-  fnames <- expand_names(indfile)
-  if(!file.exists(fnames$cache)) stop(paste(fnames$data, "is missing from the cache"))
-  message("note ", fnames$var)
-  
-  # write the indicator file
-  sc_indicate(fnames$indicator, data_file=fnames$cache)
-  invisible(NULL)
-}
```
aappling-usgs: there's a bunch of outdated proof-of-concept stuff in here. scipiper does a lot of this differently now.

R/loop_tasks.R
jread-usgs: Yes, I realized that there may actually be cases when we want verbose messages but not the progress bar after all. This is a good fix, but we might want to revisit having 3 (or 4?) message output options, including silent.

R/loop_tasks.R
jread-usgs: Yes, now that I step back from it, I think I agree. I can see a case for the current implementation, but it isn't as strong or as intuitive than changing it to your proposal.

inst/extdata/sharedcache/demo.R
jread-usgs: Good to know. I started playing around in here to find an example outside of the vignette and was scratching my head...

inst/extdata/tasks_demo/tasks_demo.R
```diff
@@ -0,0 +1,65 @@
...
+    message(sprintf("   - processing %s, resource_C=%s", task, resource_C))
+  }
+  if(rnorm(1) < 0.1) stop("processing failed")
+  return(sprintf('PROCESSED_%s', task))
+}
+visualize <- function(prep, capital) {
+  if(isTRUE(getOption('scipiper.test_verbose'))) {
+    message(sprintf("   - visualizing with prep=%s, capital=%s", prep, capital))
+  }
+  if(rnorm(1) < 0.1) stop("visualizing failed")
```
jread-usgs: when you set a seed, does this succeed or fail every time in loop tasks? I was running into that on my example that I set up to test `progress`, but I didn't play around much w/ it after I rand into it. Perhaps it was where I was setting the seed in the code.

jread-usgs: and actually, I was using `sample()` instead of `rnorm()`...not sure if that matters

tests/testthat/test-loop_tasks.R
```diff
@@ -0,0 +1,70 @@
...
+  # with verbose=FALSE, should see just one progress bar per loop attempt (shows
+  # up in output as a consecutive series of written-over progress bars with just
+  # one that's not written over by a \r)
+  set.seed(100)
+  options('scipiper.test_verbose'=NULL)
+  output <- capture_messages(scmake('models.ind'))
+  pb_lines <- grep('\\[=*>*-*\\]', output)
+  r_lines <- grep('\r', output)
+  final_pb_lines <- output[pb_lines[-which((pb_lines+1) %in% r_lines)]]
+  expect_equal(length(final_pb_lines), 1)
```
jread-usgs: clever test

R/loop_tasks.R
aappling-usgs: Agreed...I thought about pushing for that as I was struggling through this today, but this formatting change actually gets us pretty close, and I liked your suggestion to keep this simple if we can. That said, adding verbosity levels wouldn't be too tough, either, and could clean it up even more. I suggest that we keep this possibility in mind and then wait on it until it feels urgent (and then if it does, do it!).

inst/extdata/sharedcache/demo.R
aappling-usgs: Yeah, the old tests weren't so last year :stuck_out_tongue_closed_eyes: . scipiper does have a fair amount of tech debt - we're doing a lot of testing and repairing and augmenting on the fly, in conjunction with application development, but the lack of vignettes and formal tests may catch up with us, as could the fact that there is no Travis connection and even local automated Checks and Tests fail miserably right now.

...that said, I think we should continue to balance investment in scipiper with the understanding that our tools could change beneath our feet in not too long (cough cough drake), which is why I waited until today to work on these tests.

R/loop_tasks.R
aappling-usgs: OK, good, I think I can pop that into this PR.

inst/extdata/tasks_demo/tasks_demo.R
aappling-usgs: I think both `sample` and `rnorm` use the same random number generator state determined by `set.seed`. As far as I could tell, setting the seed was working for making the failures/successes reproducible here. But I couldn't get R CMD check or Ctrl-Shift-T (devtools::test_somethingsomething) to work at all with this set of tests, for completely separate reasons, and I'm not sure whether Travis/linux would produce the same sequence of successes and failures.

jread-usgs: makes sense. Whatever was happening to me was unexpected, so I think there was something else going on.

PR #82, opened 2018-12-18
tryCatch to still update build files if there is an error

### LINKED ISSUE #47: update build files even when remake::make fails partway through
if scmake successfully builds some files with `make` but is interrupted or fails partway through, the build/status files don't get built, which means the overall status database is in an inconsistent state. this shouldn't happen.

wdwatkins:
fixes #47

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -35,5 +35,5 @@ BugReports: https://github.com/USGS-R/scipiper/issues
 Encoding: UTF-8
 LazyData: true
 Roxygen: list(markdown = TRUE)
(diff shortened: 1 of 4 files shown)


aappling-usgs:
Cool! How did you verify/test that this is working?

wdwatkins:
I added this target to the remake.yml in the `tasks_demo` example:
```

aappling-usgs:
Ahh, great. Are you able to check in a `testthat` test, too?

aappling-usgs:
Awesome! First merged PR of the sprint :confetti_ball:

PR #83, opened 2018-12-18
travis yml

wdwatkins:


wdwatkins:
self-merging since this isn't affecting the actual code base

PR #84, opened 2018-12-18
version bump to try to kick off travis

wdwatkins:


.travis.yml
```diff
@@ -23,8 +23,8 @@ addons:
 env:
   global:
   - R_BUILD_ARGS="--no-build-vignettes --no-manual"
-- R_CHECK_ARGS="--no-build-vignettes --no-manual --as-cran"
-- _R_CHECK_FORCE_SUGGESTS_=false
+  - R_CHECK_ARGS="--no-build-vignettes --no-manual --as-cran"
+  - _R_CHECK_FORCE_SUGGESTS_=false
 
 warnings_are_errors: true
```
ldecicco-USGS: check below...you don't want `travis_wait 20 R CMD check dataRetrieval*tar.gz`

wdwatkins: 😬

ldecicco-USGS: up above there are indentation errors too in the matrix:

aappling-usgs:
I'm looking at this from the angle of fixing the vignettes (though I agree that being able to skip building them would be nice). The Travis-CI  error is
```
Quitting from lines 176-179 (task_plans.Rmd) 
Error: processing vignette 'task_plans.Rmd' failed with diagnostics:
All tries are exhausted, but 3 tasks remain
Execution halted
```
Unfortunately, this vignette already builds reliably without error on my computer with `devtools::build_vignettes()`, so it's one of those environment configuration things. I can reproduce the problem locally with `R CMD check` and so will start there.

For the other angle, of skipping building the vignettes on Travis, are you able to configure your local build to successfully skip the vignettes, and if so, does it work to use those same arguments on Travis?

wdwatkins:
`--no-build-vignettes` works locally, but not on Travis.  I think the flags must be different in Linux R.  I'm going to experiment on my VM and see if I can figure it out.

aappling-usgs:
What a mess. Good luck!

wdwatkins:
I'm giving up on this for now.  `--no-build-vignettes` works on my Ubuntu VM.  I think there must be an issue with how Travis translates the yaml into system commands.

wdwatkins:
Travis is running now, so I think we can go ahead an merge this

Issue #52, opened 2018-06-15
support combiners in task tables


jread-usgs:
`sc_indicate` is great for a closing off a series of tasks w/ a single target. 

For the gridded data builds from NLDAS, I find myself needing to write a single summary file after all the tasks are done. In this case, it is `'0_config/in/nldas_cells.rds'`, which is a cheater file that provides an overview of existing data that informs the original data pulls. To include this in remake directly is a cyclical reference no-no, but it is important nonetheless, and I think you understand the pattern I am using. 

Anyhow, I am curious about the best way to connect the rebuild of that file to `sc_indicate`. I could: 
1) have a separate remake file that has a target that depends on the indicator file that gathers all of the targets in the big task plan. This would have the side-effect of updating that .rds file w/ the now current data
2) sc_indicate could accept a function as an optional argument, and could run that function as a final step to executing its normal duties. In this case I would do something like `sc_indicate(ind_file = "example.ind", exit_function = summarize_nldas_cells(cells_dir, '0_config/in/nldas_cells.rds'))` but I see that would be definitely more complex

I think we have kicked around some thoughts related to a more explicit "combine" method available to `scipiper`. This is one potential use-case where such a method would be really helpful.

jread-usgs:
Another use-case for this:

I've got a bunch of netcdf files, and I want to spawn a bunch of feather files from those netcdf files. Since there is significant overhead in opening and closing the netcdf files, I would like to

aappling-usgs:
Thanks for these thoughts. I appreciate the specific use case and agree that something is needed here. Rather than overloading `sc_indicate`, I'd propose that we allow additional combine operations as part of `create_task_makefile`, which could then still be responsible for producing a single "I'm all done!" indicator with `sc_indicate` but which could also permit any number of combination operations that depend on the completion of intermediate steps.

`create_task_makefile(task_plan, combine_plan[new], [...other args...])`

jread-usgs:
I'm working through the code a bit and re-familiarizing myself w/ how `scipiper` "completes" the task tables. 

Currently, we can tie together the final steps in each task as depends of a single indicator file (the default), or we can `hash_dependencies()` to keep the result as an object in R (in this case a vector of the hashes). These are set by `ind_complete = TRUE` and `FALSE`, respectively. 

[...]

PR #86, opened 2018-12-19
initial ideas for combine #52

jread-usgs:
initial PR for comment


R/create_task_makefile.R
```diff
@@ -82,23 +83,43 @@ create_task_makefile <- function(
...
+    tasks[[task]]$steps <- unname(tasks[[task]]$steps)
+    for(step in seq_along(tasks[[task]]$steps)) {
+      # add a logical for has_depends
+      tasks[[task]]$steps[[step]]$has_depends <- length(tasks[[task]]$steps[[step]]$depends) > 0
+    }
+  }
+  
   job <- list(
-    target_name = job_target,
+    target_name = target_name,
```
aappling-usgs: I think there's still room for a non-combiner summary job_target (which stores the file names and hashes of the final step from each task), so I'd rather not repurpose that existing summary target here (or allow the user to choose its name; it's fixed to enforce uniformity across projects). In particular, a job target is useful if you don't need a combiner or haven't written it yet, or if you want to commit that job .ind file so that you can git diff any changes to the output of individual steps.

Could we instead allow the user to specify the target_name[s] of the combiner[s] as other targets, to complement the `job_target` if `ind_complete=TRUE` and to replace the `job_target` if `ind_complete=FALSE`?

R/create_task_makefile.R
```diff
@@ -82,23 +83,43 @@ create_task_makefile <- function(
...
     # even though target_name is an object (not file), job_command should write
     # to ind_file - again so the calling remake file can use
     # ind_file as its target
     command = {
+      if(!is.null(finalize_fun)){
+        if(isTRUE(ind_complete) | is.na(ind_complete)) {
+          message('ind_complete=', ind_complete, ' is incompatable with ', finalize_fun, '(). Using FALSE')
```
aappling-usgs: I don't think `ind_complete` and the output of `finalize_fun` should be forced not to coexist, though I could see a warning or message encouraging the user not to bother with `ind_complete=TRUE` if they're providing one or more combiners.

R/create_task_makefile.R
```diff
@@ -82,23 +83,43 @@ create_task_makefile <- function(
...
+          ind_complete <- FALSE
+          # and will need to skip hashing...
+        }
+        sprintf("%s(target_name = %s,\n)", finalize_fun, target_name) # and more w/ ...
```
aappling-usgs: I'm wondering - do we want to make it possible to run the combiner before all the tasks are complete? This is often useful while developing code (e.g., try merging a week of .pngs into a .gif before building the whole year, or work out the code for combining .netcdf files before downloading all the .netcdf data). There are a few ways a user could make that happen without modifying this code, but they're each a little bit clunky, and there might be a clever way to implement this combiner so that it only operates on the completed steps, whichever those are...food for thought.

R/create_task_makefile.R
jread-usgs: yes, I agree. Will change

R/create_task_makefile.R
jread-usgs: Part of this is temporary as I work through the implementation. But I am still wondering about the need to indicate the final single output within this yaml, vs sc_indicate from the parent yaml.

R/create_task_makefile.R
jread-usgs: Not sure - my initial response would be that they should make the dev version of the makefile a truncated number of tasks. 

or, you could edit your combiner in dev mode:
```r
my_combine <- function(target_name, ...){
   files <- c(...)[1:10] 
   message("I am skipping the rest of the files!!!")
   # do things
   write_feather(target_name, stuff)
}
```

R/create_task_makefile.R
aappling-usgs: Agreed that there's no need to produce an .ind file for a combiner output; a target name specific to that combiner, of the user's choosing, would be fine.

I do think we still want an option to produce the .ind file for summary output, because that's the one that can put together a single list of filenames and hashes for the final steps of all of the tasks. The target that's implemented for that summary output does write an .ind file but uses an object name / symbol as the recipe target name (for the reason noted at https://github.com/USGS-R/scipiper/issues/52#issuecomment-448623127)

R/create_task_makefile.R
aappling-usgs: Editing the combiner wouldn't work because remake would see all the ...s and wouldn't run the combiner until they'd all been built.

So far today I agree with your initial response. I think I've had reasons not to want to do it that way in the past, but I can't dredge them up today at all.

jread-usgs: ahh yep, you are right.

jread-usgs:
example:

```r
task_config <- data.frame(
    id=c('AZ','CO','CA'),
    capital=c('Phoeniz','Denver','Sacramento')
)
step1 <- create_task_step(
    step_name = 'prep',
    target = function(task_name, step_name, ...) {
        sprintf('%s_%s', task_name, step_name)
    },
    depends = c('A','B'),
    command = "process(target_name, I('C'))"
)
step2 <- create_task_step(
    step_name = 'plot',
    command = function(target_name, task_name, ...) {
        capital <- task_config[task_config$id == task_name, 'capital']
        sprintf('visualize(\'%s\', \'%s\')', task_name, capital)
    }
)
step3 <- create_task_step('report')
task_plan <- create_task_plan(c('AZ','CA','CO'), list(step1, step2, step3),
                              final_steps='report', ind_dir='states/log')
```

now use the normal task_makefile:
```r
task_config <- data.frame(
    id=c('AZ','CO','CA'),
    capital=c('Phoeniz','Denver','Sacramento')
)
step1 <- create_task_step(
    step_name = 'prep',
    target = function(task_name, step_name, ...) {
        sprintf('%s_%s', task_name, step_name)
    },
    depends = c('A','B'),
    command = "process(target_name, I('C'))"
)
step2 <- create_task_step(
    step_name = 'plot',
    command = function(target_name, task_name, ...) {
        capital <- task_config[task_config$id == task_name, 'capital']
        sprintf('visualize(\'%s\', \'%s\')', task_name, capital)
    }
)
step3 <- create_task_step('report')
task_plan <- create_task_plan(c('AZ','CA','CO'), list(step1, step2, step3),
                              final_steps='report', ind_dir='states/log')
task_makefile <- create_task_makefile(
    task_plan, makefile=file.path(tempdir(), 'states.yml'),
    file_extensions=c('ind'), packages='mda.streams')
```
```yaml
# Do not edit - automatically generated
# from the task_makefile.mustache template
# by create_task_makefile()

target_default: states
    depends:
      - states_ind

packages:
  - mda.streams

file_extensions:
  - "ind"

targets:
  # --- AZ --- #
  
  AZ_prep:
    command: process(target_name, I('C'))
    depends:
      - A
      - B

  AZ_plot:
    command: visualize('AZ', 'Phoeniz')

  AZ_report:
    command: report(I('AZ'))

  states/log/AZ.ind:
    command: sc_indicate(target_name)
    depends:
      - AZ_report

  # --- CA --- #
  
  CA_prep:
    command: process(target_name, I('C'))
    depends:
      - A
      - B

  CA_plot:
    command: visualize('CA', 'Sacramento')

  CA_report:
    command: report(I('CA'))

  states/log/CA.ind:
    command: sc_indicate(target_name)
    depends:
      - CA_report

  # --- CO --- #
  
  CO_prep:
    command: process(target_name, I('C'))
    depends:
      - A
      - B

  CO_plot:
    command: visualize('CO', 'Denver')

  CO_report:
    command: report(I('CO'))

  states/log/CO.ind:
    command: sc_indicate(target_name)
    depends:
      - CO_report

  # --- Overall job --- #

  states_ind:
    command: sc_indicate(I('states/log/states.ind'))
```

or instead, specify a combiner function that is available via the sources:

```r
task_makefile <- create_task_makefile(
    task_plan, makefile=file.path(tempdir(), 'states.yml'),
    file_extensions=c('ind'), packages='mda.streams', ind_complete=FALSE, finalize_fun='my_combiner', target_name='file_out.csv')
```
```yaml
# Do not edit - automatically generated
# from the task_makefile.mustache template
# by create_task_makefile()

target_default: states
    depends:
      - file_out.csv

packages:
  - mda.streams

file_extensions:
  - "ind"

targets:
  # --- AZ --- #
  
  AZ_prep:
    command: process(target_name, I('C'))
    depends:
      - A
      - B

  AZ_plot:
    command: visualize('AZ', 'Phoeniz')

  AZ_report:
    command: report(I('AZ'))

  states/log/AZ.ind:
    command: sc_indicate(target_name)
    depends:
      - AZ_report

  # --- CA --- #
  
  CA_prep:
    command: process(target_name, I('C'))
    depends:
      - A
      - B

  CA_plot:
    command: visualize('CA', 'Sacramento')

  CA_report:
    command: report(I('CA'))

  states/log/CA.ind:
    command: sc_indicate(target_name)
    depends:
      - CA_report

  # --- CO --- #
  
  CO_prep:
    command: process(target_name, I('C'))
    depends:
      - A
      - B

  CO_plot:
    command: visualize('CO', 'Denver')

  CO_report:
    command: report(I('CO'))

  states/log/CO.ind:
    command: sc_indicate(target_name)
    depends:
      - CO_report

  # --- Overall job --- #

  file_out.csv:
    command: my_combiner(target_name = file_out.csv,
      states/log/AZ.ind,
      states/log/CA.ind,
      states/log/CO.ind)

```

limitations 
 - this includes the _possibility_ of multiple "final" targets, because `length(job)` can be > 1 and it will work, but it isn't tied to the current arguments (you'd have to change the codes to actually use this though)
 - the format for a combiner function is limited to `function(target_name, ...)`, meaning only filenames can be used. It would be nice to support `function(...)` that return an object by detecting if target_name is an actual file with `remake:::target_is_file()`
 - this is rigid and does not allow you to combine any steps other than `final` ones.

aappling-usgs:
Looks promising. See discussion in slack about hanging `depends:` off of `target_default:` and maybe sticking with `states` as the target name for the summary .ind file, using `states_all` or similar as the new default target name.

> this includes the possibility of multiple "final" targets, because length(job) can be > 1 and it will work, but it isn't tied to the current arguments (you'd have to change the codes to actually use this though)

What is `job`?

> the format for a combiner function is limited to function(target_name, ...), meaning only filenames can be used. It would be nice to support function(...) that return an object by detecting if target_name is an actual file with remake:::target_is_file()

Couldn't you simply ignore that first argument for a combiner that returns an object? I agree that function(...) would be a little bit cleaner, so just depends on how much time it'd take to support the latter vs just accepting the slightly less clean workaround

> this is rigid and does not allow you to combine any steps other than final ones.

OK

aappling-usgs:
> Couldn't you simply ignore that first argument for a combiner that returns an object? I agree that function(...) would be a little bit cleaner, so just depends on how much time it'd take to support the latter vs just accepting the slightly less clean workaround

> I think remake won't support that, unless I wrapped used target_name = I({{target_name}}). Since I'd need to sniff out if it was a file or not to do that, I'd rather skip that arg when it is an object

Oh, OK, I didn't realize you couldn't use target_name for object targets.

> What is job?

> job is this internal list guy. I modified it in my PR, but it still represents the same thing

I see it now. 

> And how hard would it be to fully support multiple final targets?

> It shouldn't be too bad, but I had some questions before progressing. Main one is whether or not hash_depends is a necessary option to keep in here or not. I think it could be a type of combiner and thus I'd want to take out the special block of code that includes it and instead have people specify it in the same way they specify other combiners.

Cool idea! I'm game.

jread-usgs:
> Oh, OK, I didn't realize you couldn't use target_name for object targets.

I'm going to double check that...I realize the syntax I am coding in is `target_name = %s`, when I think I just need `target_name`, which may work for the objects and we'd be :+1:

jread-usgs:
Added new code per our discussion. 

```r
task_makefile <- create_task_makefile(
    task_plan, makefile=file.path(tempdir(), 'states.yml'),
    file_extensions=c('ind'), packages='mda.streams', finalize_fun=c('my_combiner','other_function'), target_name=c('r_object', 'other_file.csv'))
```
creates
```yaml
# Do not edit - automatically generated
# from the task_makefile.mustache template
# by create_task_makefile()

target_default: states

packages:
  - mda.streams

file_extensions:
  - "ind"

targets:
  states:
    depends:
      - r_object_promise
      - other_file.csv_promise

  # --- AZ --- #
  
  AZ_prep:
    command: process(target_name, I('C'))
    depends:
      - A
      - B

  AZ_plot:
    command: visualize('AZ', 'Phoeniz')

  AZ_report:
    command: report(I('AZ'))

  states/log/AZ.ind:
    command: sc_indicate(target_name)
    depends:
      - AZ_report

  # --- CA --- #
  
  CA_prep:
    command: process(target_name, I('C'))
    depends:
      - A
      - B

  CA_plot:
    command: visualize('CA', 'Sacramento')

  CA_report:
    command: report(I('CA'))

  states/log/CA.ind:
    command: sc_indicate(target_name)
    depends:
      - CA_report

  # --- CO --- #
  
  CO_prep:
    command: process(target_name, I('C'))
    depends:
      - A
      - B

  CO_plot:
    command: visualize('CO', 'Denver')

  CO_report:
    command: report(I('CO'))

  states/log/CO.ind:
    command: sc_indicate(target_name)
    depends:
      - CO_report

  # --- Overall job --- #

  r_object_promise:
    command: my_combiner(I(r_object),
      states/log/AZ.ind,
      states/log/CA.ind,
      states/log/CO.ind)
    
  other_file.csv_promise:
    command: other_function(I(other_file.csv),
      states/log/AZ.ind,
      states/log/CA.ind,
      states/log/CO.ind)
```

aappling-usgs:
Cool, and then for the record, we said we'd probably reference the combiner outputs in the parent remake file something like this, right?

```yml
r_object:
  scmake(I('r_object_promise'), remake_file='states.yml')

other_file.csv:
  scmake(remake_file='states.yml')
```

aappling-usgs:
Does it work to use sc_indicate as a combiner? I just realized sc_indicate accepts the information about data files as a single `data_file` argument rather than in its `...`s, so we might need a very lightweight wrapper that converts from `...` to a vector.

jread-usgs:
You are right...this doesn't work w/ `sc_indicate` with the current pattern

jread-usgs:
@aappling-usgs curious if you see the need for flexibility whether or not we use the `_promise` (and the corresponding `I({data_file})`)

I could see an additional arg, `as_promises`, or something like that that is `TRUE` by default. This way, if you are using the tools simply to generate a templated task table, and have no linkage to a parent remakefile, you avoid the promises elements to this.

jread-usgs:
to do: 
 - [ ] add `as_promises = TRUE` as arg
 - [ ] add `packages = `scipiper` as default arg (?)
 - [ ] allow `finalize_fun` to be `NULL`, which would tie all of the final steps to the default target as depends. 
 - [ ] use code in #90 to simplify
 - [ ] check for file vs object, determine whether to pass in `target_name`

jread-usgs:
With ☝️ updates, we can:
 - specify `makefile = NULL` to return the rendered string (instead of writing to file)
 - specify `finalize_fun = NULL` to skip the combiner and make all of the final step targets depends on the default/all target
 - there is a default `combine_to_ind` function for `finalize_fun` that is a wrapper to `sc_indicate`
 - the quoting of files and non-quoting of R objects works as expected.

aappling-usgs:
> make all of the final step targets depends on the default/all target
Did you mean "dependencies of" or (equivalently) "make the default/all target depend on all of the final step targets"?

R/create_task_makefile.R
```diff
@@ -25,12 +21,16 @@
...
 #' @param template_file character name of the mustache template to render into
 #'   `makefile`. The default is recommended
-#' @return the file name of the makefile that was created (can be displayed with
-#'   `cat(readLines(makefile), sep="\n")`).
+#' @param target_name a file path (or vector of file paths) specifying final target(s) 
+#'   to be created by `finalize_fun`
+#' @param finalize_fun a string function name (or vector of function names, of equal 
+#'   length to `target_name`). Using NULL for `finalize_fun` will turn the 
+#'   final steps from the task plan into depends for the default/all target.
+#' @param as_promises ... details here...
```
aappling-usgs: remember to fill in these details

R/create_task_makefile.R
```diff
@@ -59,63 +59,48 @@
...
   # ind_file as a target (even though this target is the one responsible for
   # actually writing to ind_file)
-  job_target <- tools::file_path_sans_ext(basename(makefile))
+  if (is.null(makefile) & missing(target_name)){
+    stop('target_name must be specified when makefile is NULL')
+  }
+  # default is to use the makefile naming to assign job name
+  # when makefile isn't used (string output), we use the target_name to create default job name
+  job_name <- ifelse(is.null(makefile), 
+                     paste0(tools::file_path_sans_ext(basename(target_name)), '_all'),
```
aappling-usgs: What if you want to output the makefile as string but you don't want any combiners (or have multiple combiners)? Could the job_name be something bland like `all_tasks`?

R/create_task_makefile.R
```diff
@@ -59,63 +59,48 @@
...
+                     tools::file_path_sans_ext(basename(makefile))) 
+  
+  if(missing(target_name)){
+    ind_ext <- getOption("scipiper.ind_ext") # put this here because if you want to control the extension, specify the target_name
+    ind_dir <- attr(task_plan, 'ind_dir') # if you want to control the dir over default, specify the target_name
+    target_name <- file.path(ind_dir, as_ind_file(job_name, ind_ext))
+  }
+  
+  if (!is.null(finalize_fun) & length(target_name) != length(finalize_fun)){
+    stop('when specifying function names for `finalize_fun`, an equal number of `target_name`(s) need to be specified', call. = FALSE)
```
aappling-usgs: Is `call. = FALSE` necessary? I usually find the call information to be informative, as long as it doesn't overwhelm all the other information.

R/create_task_makefile.R
```diff
@@ -127,6 +112,46 @@ create_task_makefile <- function(
...
+          # hide the object/file changes as input when using `promises`
+          if (target_is_file){
+            ifelse(as_promises, {
+              sprintf("%s(I('%s'),\n      %s)", finalize_fun[i], target_name[i], to_combine)
+            }, {
+              sprintf("%s(target_name,\n      %s)", finalize_fun[i], to_combine)
+            })
+          } else {
+            combine_str <- "%s(\n      %s)"
+            sprintf(combine_str, finalize_fun[i], to_combine) # use PR #90 when it is merged
```
aappling-usgs: Could take this comment out now. For the record, I think if you really wanted to use psprintf, you could do:
```
do.call(
  scipiper::psprintf,
  c(
    list('%s('=finalize_fun),
    sprintf('%s,', to_combine[-length(to_combine)]),
    sprintf('%s)', to_combine[length(to_combine)])
  )
)
```
but I like your way better.

R/create_task_makefile.R
jread-usgs: removing :+1:

R/create_task_makefile.R
jread-usgs: Good idea, will update to that

tests/testthat/test-create_task_makefile.R
```diff
@@ -0,0 +1,110 @@
...
+    task_plan, makefile=file.path(tempdir(), 'states.yml'), target_name = 'my_states',
+    file_extensions=c('ind'), packages=c('scipiper','mda.streams'), as_promises = FALSE, 
+    finalize_fun = c('combine_these','combine_those')))
+})
+
+test_that("task_makefiles with multiple combiners", {
+  task_makefile <- create_task_makefile(
+    task_plan, makefile=file.path(tempdir(), 'states.yml'), target_name = c('these_states','those_states'),
+    file_extensions=c('ind'), packages=c('scipiper','mda.streams'), as_promises = FALSE, 
+    finalize_fun = c('combine_these','combine_those'))
```
aappling-usgs: Running this command generates these messages:
```
run all tasks with
these_states:
  command: scmake(remake_file='C:\Users\aappling\AppData\Local\Temp\1\RtmpMb8XgC/states.yml')
run all tasks with
those_states:
  command: scmake(remake_file='C:\Users\aappling\AppData\Local\Temp\1\RtmpMb8XgC/states.yml')
```

On the one hand, maybe that's fine because (1) both suggestions are true and (2) does anybody read these messages anyway?

On the other hand, it'd be a little nicer if the message for how to "run all tasks" didn't recommend scmaking the same target twice, wouldn't it? Something like:
```
Run all tasks and finalizers with:
these_states:
  command: scmake(I('these_states_promise'), remake_file='C:/Users/aappling/Documents/states.yml')
those_states:
  command: scmake(I('those_states_promise'), remake_file='C:/Users/aappling/Documents/states.yml')
```

I _think_ I've implemented something workable at https://github.com/aappling-usgs/scipiper/commit/bd952c530d6d224290193269c35e1a7043a65122, but I haven't tested it for all the possible combinations of target_name, as_promises, and whatever else is relevant. We could push this to a separate issue if needed.

tests/testthat/test-create_task_makefile.R
```diff
@@ -0,0 +1,110 @@
...
+  expect_equal(tail(remake_list$targets, 1)[[1]]$command, 
+               "combine_to_ind(I('states/log/all_tasks.ind'), 'states/log/AZ.ind', 'states/log/CA.ind', 'states/log/CO.ind')")
+})
+
+test_that("can create task_makefile with named target", {
+  task_makefile <- create_task_makefile(
+    task_plan, makefile=NULL, target_name = 'my_states',
+    file_extensions=c('ind'), packages=c('scipiper','mda.streams'))
+  remake_list <- yaml::yaml.load(task_makefile)
+  expect_equal(tail(names(remake_list$targets), 1), "my_states_promise")
```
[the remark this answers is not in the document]
jread-usgs: Geez, that was short-sighted of me. I like your idea, but it brings up a tricky part of the implementation. If we use the last arg as the named file path for `combine_to_ind` to pass `ind_file` to `sc_indicate`, we would then need to have consistency in arg naming of other combiner functions. Something like this:

```r
combine_to_ind(..., filepath){
  # bla bla
}
```
or more generic as
`combine_to_ind(..., x)`

My thinking on this originally (let me know if this no longer holds water) was to have two types of combiner functions, ones that write files and ones that return objects. The file writers take a file as the first arg and `...` for the rest of the objects/files to be part of the combiner operations. The object creators are the same, but have no named arg (only `...`). So me about 2 months ago would say, "well, `combine_to_ind` is a file writer combiner, and you are using it incorrectly, you must want `combine_to_hash`..." But the behavior you are pointing out above is kind of dangerous, because you would end up unknowingly re-writing the contents of 'states/log/AZ.ind' to reflect the combined hashes or the other two files...which isn't good.

tests/testthat/test-create_task_makefile.R
```diff
@@ -0,0 +1,110 @@
...
+  expect_equal(head(names(remake_list$targets), 1), "my_states_all")
```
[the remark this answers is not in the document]
jread-usgs: Not good. Good catch. Fixed

tests/testthat/test-create_task_makefile.R
jread-usgs: I punted on this one, as handling all of those different combinations (e.g., targets, promises, etc) seemed time consuming and error prone, and I wasn't sure we wouldn't want to change the implementation after review or after playing with it.

jread-usgs: Ok - I tried to PR your branch to mine, but you have pitfalls vignette commits in there, so I stole your code for the message instead...lost attribution! 👎 but it looks good to me 💯

jread-usgs:
I updated the code per peer review. Tests run fine locally, but one test seems to fail when I run `devtools::check()`:
```
Running ‘testthat.R’ ERROR
Running the tests in ‘tests/testthat.R’ failed.
Last 13 lines of output:
         stop(e)
     }, "4 simultaneous processes spawned", quote(.check_ncores(length(names))))
  24: h(simpleError(msg, call))
  25: stop(e)
  26: (function (e) 
     {
         update_build_files(target_names = target_names, remake_file = remake_file, status_pre = status_pre)
         stop(e)
     })(structure(list(message = "4 simultaneous processes spawned", call = .check_ncores(length(names))), .Names = c("message", 
     "call"), class = c("simpleError", "error", "condition")))
```
Which I didn't think had impacted. Odd

tests/testthat/test-create_task_makefile.R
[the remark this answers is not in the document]
aappling-usgs: Just wrote four different replies to this thread. Here's my final response (for now :smile:):

For non-default finalizers, I think the user just has to be responsible for creating an appropriate match between their `final_targets` and `finalize_funs`, and I don't see any reliable ways for us to check on the user's behalf within `create_task_makefile`. Therefore there's no huge advantage to moving the ind_file argument to the end, and I agree that keeping it first has the nice property of allowing the user to define their own argument name for the output file.

However, `combine_to_ind` is special because it's the default. If `final_targets` is not supplied, the current code chooses a filename. I think it would be reasonable and useful for `create_task_makefile` to check and throw an error if `length(finalize_fun) == length(final_targets) == 1` and `finalize_fun=='combine_to_ind'` and yet `!remake:::target_is_file(final_targets, file_extensions = pos_extensions)`.

To specify a `final_target` that's an object, the user (and the tests) will therefore have to also specify a different `finalize_fun`, anything except `combine_to_ind`. I think that'd be fine/good.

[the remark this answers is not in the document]
aappling-usgs: What do you think @jread-usgs - add a check with stop() in create_task_makefile and then modify the above test to expect or not trigger that stop()? Or did you already do this?

aappling-usgs:
Hi @jread-usgs , I think I'm waiting on you for a decision on https://github.com/USGS-R/scipiper/pull/86#discussion_r255132174. I don't see that local `devtools::check()` error showing up in Travis and so am inclined to ignore it for the moment...but of course if we see it again in the future we should probably figure it out. After that, I think this PR is good to merge, and after a couple of weeks of people testing it on this separate branch, we should pull it into the USGS-R master branch.

tests/testthat/test-create_task_makefile.R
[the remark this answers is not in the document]
jread-usgs: Great ideas. I forgot this was left unanswered by me. I agree w/ your expectation for the defaults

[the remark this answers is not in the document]
jread-usgs: Instead of checking for length(final_targets) == 1, etc. I checked that the finalize_funs[i] wasn't `combine_to_ind` when the final_targets[i] was an object. This covers your case, but also a few other cases too and was simpler to implement and simpler for future people to understand the code at that spot.

aappling-usgs:
Should be good to go with commit b8ea8ac...I'll just wait for the tests to pass and then merge :+1:

tests/testthat/test-create_task_makefile.R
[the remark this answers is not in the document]
aappling-usgs: Yeah, nice solution!

Issue #60, opened 2018-07-24
force=TRUE option for loop_tasks


aappling-usgs:
Splitting this issue off from #4

It would be swell if loop_tasks, or scmake called on a recipe with command=loop_tasks, could accept a force=TRUE argument in which case it would delete some/all tasks and/or steps before attempting to rebuild them.

Maybe even while setting aside the deleted files during the build so they can be replaced if the build encounters errors (but that's optional icing on the cake).

And it would be triply swell if that command also gave us the option to choose which tasks and steps to force the rebuild for, and which to leave alone. Maybe an additional argument that looks something like force_steps=c(1,4,5) or force_tasks=c(3,6)?

aappling-usgs:
The first and third ideas from the issue description have been resolved in #87. What's left:

* Set aside the deleted files during the build so they can be replaced if the build encounters errors (would be nice, but not supported by the more general `scmake(..., force=TRUE)` either, so not super high priority)

aappling-usgs:
Let's also tack on:
* Support deletion of `_promise` targets for combiners during `force=TRUE` (https://github.com/USGS-R/scipiper/issues/52#issuecomment-464836963)

PR #87, opened 2018-12-19
first part of force=TRUE for loop_tasks

lindsaycarr:
Starting with just deleting the files; will look into the other parts of #60 later. Here's what I am running to test this functionality (existing tests are failing, but can't get into fixing right now)

```
# setup
setwd("D:/LRCData/R/scipiper/inst/extdata/tasks_demo/") # bad, but just temporary to test how this runs
task_plan <- scmake("task_plan")
scmake('models.yml')

# build the loop one time
loop_tasks(task_plan, 'models.yml')

# verify that files exist
all(file.exists("tmp/AZ.ind", "tmp/CA.ind", "tmp/CO.ind", "tmp/models.ind"))

# now use force=TRUE to delete the appropriate files and rebuild
loop_tasks(task_plan, 'models.yml', force = TRUE)
```


aappling-usgs:
You should be able to pull upstream master now and get it down to 1 error, all but this one:
```
test-loop_tasks.R:46: failure: can run loop_tasks to completion even when tasks fail sometimes
length(final_pb_lines) not equal to 1.
1/1 mismatches
[1] 0 - 1 == -1
```

inst/extdata/tasks_demo/tasks_demo.R
```diff
@@ -51,7 +51,7 @@ demo_task_plan <- function(task_config) {
     task_names=c('AZ','CA','CO'),
     task_steps=list(step1, step2, step3),
     final_steps='report',
-    ind_dir='.',
+    ind_dir='tmp',
```
aappling-usgs: why `tmp` instead of `log` (or no change, for this very simple demo)?

tests/testthat/test-loop_tasks.R
```diff
@@ -20,8 +20,8 @@ test_that("can run loop_tasks to completion even when tasks fail sometimes", {
   # with verbose=TRUE, should see errors but also completion
   set.seed(100)
   options('scipiper.test_verbose'=TRUE)
-  output <- capture_messages(scmake('models.ind'))
-  expect_true(all(file.exists('models.ind','AZ.ind','CA.ind','CO.ind')), info='all files get created')
+  output <- capture_messages(scmake('tmp/models.ind'))
+  expect_true(all(file.exists('tmp/models.ind','AZ.ind','CA.ind','CO.ind')), info='all files get created')
```
aappling-usgs: Now that `ind_dir='tmp'` in the `create_task_plan()` call above, I'm guessing that `AZ.ind`, etc. get written to `tmp/AZ.ind`, etc. - is that the case? If so then this test should change to correspond to the new `ind_dir`.

inst/extdata/tasks_demo/tasks_demo.R
lindsaycarr: I couldn't get "." to work. It couldn't ever find the files because "." went to my home directory. I can make it "log" instead!

tests/testthat/test-loop_tasks.R
lindsaycarr: oh, I thought I made that change. Maybe I didn't push to master...

inst/extdata/tasks_demo/tasks_demo.R
aappling-usgs: Interesting. I haven't had a problem with '.' and have been running this a lot this week. We did do some debugging on David's computer, where one time he ran scmake() from within the inst/extdata/tasks_demo folder, which added the .remake folder and other files, which changed the behavior of his tests because everytime `setup_tasks_demo()` was called it was copying the .remake/* files into that new temp directory....doesn't sound like exactly your problem, but I wonder if there's something sorta like that going on.

lindsaycarr: Well, I was doing scmake() from within the `inst/extdata/tasks_demo` folder before the tests were working to check that my new functionality worked. Running scmake from there caused issues with the `.`, but that was all before I ran tests. Changing to "log" fixed everything for me. I don't think it had anything to do with tests because I was having these issues without even touching tests.

lindsaycarr: Maybe it's the nature of how `setwd()` works within the `setup_tasks_demo()` function, so my little workaround to just try out the demo using `setwd("inst/extdata/tasks_demo")` and then `scmake()` didn't work out with "." as planned.

lindsaycarr: Hmmm yeah, just tried running `setup_tasks_demo()` and then my code to test out if it's working with "." as the directory and it worked. I'll revert the folder stuff

aappling-usgs: :+1:

PR #90, opened 2018-12-19
psprintf helper function 

jzwart:
Initial function and documentation for `psprintf`, a helper function that pastes a bunch of `sprintf` text and variable values together, with a default separator as a new line. This function is useful for `task plan` creation, for example, when a target command requires multiple task-specific values. 

I also added a quick example to the task plan vignette. example in vignette: 
```    
psprintf(
      "visualize(",
      "plot_name = target_name,",
      "state_data = %s_prep,"=task_name,
      "state_capital = I('%s'))"=capital
      )
````
results in: 
```
visualize(
      plot_name = target_name,
      state_data = WI_prep,
      state_capital = I('Madison'))
```

### CODE REVIEW

**Comment by jzwart** [comment]
this addresses #56

File: R/psprintf.R [hunk 1]
```diff
@@ -0,0 +1,35 @@
+#' Paste together sprintf-style text and variable values
+#'
+#' Helper function to sprintf a bunch of key-value (string-variableVector) pairs,
+#' then paste them together with a good separator for contructing remake recipes.
+#'
+#' @param sep separator for the list of arguments, defaults to new line
```
jread-usgs: Add a line for: 
`#' @param \dots {explanation}` ?

Otherwise we'll get a check warning for an undocumented arg when we get in better shape w/ the package check stuff.

vignettes/task_plans.Rmd
```diff
@@ -73,15 +73,23 @@ step1 <- create_task_step(step_name = 'prep')
 
 # step2 creates recipes like this:
 # WI_plot.png:
-#   command: visualize(target_name, 'WI_prep', I('Madison'))
+#   command: visualize(
```
jread-usgs: Should we delete the commented out lines?

R/psprintf.R
jzwart: Yup I can add a description

vignettes/task_plans.Rmd
aappling-usgs: Unless they're inaccurate, I think these comments provide useful, proximate explanation of what the `create_task_step()` calls do.

jzwart: I kind of like the example of what each step will produce before it is run so I'm for keeping in the commented lines.

aappling-usgs: (and I just checked...I think they're still accurate)

R/psprintf.R
aappling-usgs: For the description, how about "arguments that are either (1) character strings to insert directly into the output or (2) named arguments, where the name is an sprintf pattern and the value is a list or vector of variables to apply to that particular sprintf pattern. After any relevant sprint evaluation, all arguments will be pasted, with the specified separator, in order to create a single character output."

R/psprintf.R
```diff
@@ -0,0 +1,36 @@
...
+#'
+#' @export
+#' @examples
+#' \dontrun{
+#' psprintf(
+#'   "plot_frame(",
+#'   "png_file = target_name,",
+#'   "date = date_%s," = "2018-01-23",
+#'   "state = %s)" = "Wisconsin")
+#' }
```
aappling-usgs: The above is a good example, and maybe the best one, but for the record (and if you think these should go in, please add them), here are a couple of others that exercise the functionality from my `...` description above:
```
cat(psprintf(
  'cmd(',
  'a = %s, b = \'%s\','=c('dog','cat'),
  'c = "rat")'
))
## cmd(
##       a = dog, b = 'cat',
##       c = "rat")
```

```
cat(psprintf(
  'cmd(',
  'a = %s, b = %0.1f,'=list('dog',7),
  'c = "rat")'
))
## cmd(
##       a = dog, b = 7.0,
##       c = "rat")
```

R/psprintf.R
```diff
@@ -0,0 +1,36 @@
...
+#'
+#' Helper function to sprintf a bunch of key-value (string-variableVector) pairs,
+#' then paste them together with a good separator for contructing remake recipes.
+#'
+#' @param ... list of sprintf arguments seperated by commas
+#' @param sep separator for the list of arguments, defaults to new line
+#'
+#' @export
+#' @examples
+#' \dontrun{
```
aappling-usgs: Is the \dontrun necessary? This example runs fine for me

R/psprintf.R
jzwart: Ah yeah that's much better

R/psprintf.R
jzwart: I will add those examples. Thanks!

R/psprintf.R
jzwart: Hmm. yeah I'm not sure why I had that in there. I'll remove the `\dontrun`

Issue #88, opened 2018-12-19
get R CMD check to succeed


aappling-usgs:
Now that David has Travis running, it'd be extra nice for the package to build and check cleanly.

aappling-usgs:
Current Travis build status (e.g., https://travis-ci.org/USGS-R/scipiper/jobs/470080828):
```
Building package
Building with: R CMD build --no-build-vignettes --no-manual --no-vignettes
11.23s$ R CMD build  .
* checking for file ‘./DESCRIPTION’ ... OK
* preparing ‘scipiper’:
* checking DESCRIPTION meta-information ... OK
* installing the package to build vignettes
* creating vignettes ... ERROR
USGS Support Package: https://owi.usgs.gov/R/packages.html#support
Attaching package: 'dplyr'
The following objects are masked from 'package:stats':
    filter, lag
The following objects are masked from 'package:base':
    intersect, setdiff, setequal, union
Quitting from lines 176-179 (task_plans.Rmd) 
Error: processing vignette 'task_plans.Rmd' failed with diagnostics:
All tries are exhausted, but 3 tasks remain
Execution halted
The command "R CMD build  ." failed and exited with 1 during .
Your build has been stopped.
```

Current local `check` status:
```
==> devtools::check(cleanup = FALSE)

Updating scipiper documentation
Loading scipiper
USGS Support Package: https://owi.usgs.gov/R/packages.html#support
Writing NAMESPACE
Writing NAMESPACE
Setting env vars --------------------------------------------------------------
CFLAGS  : -Wall -pedantic
CXXFLAGS: -Wall -pedantic
Building scipiper -------------------------------------------------------------
"C:/PROGRA~1/R/R-35~1.1/bin/x64/R" --no-site-file --no-environ --no-save  \
  --no-restore --quiet CMD build "D:\APAData\GitHub\Pipeline\scipiper"  \
  --no-resave-data --no-manual 

* checking for file 'D:\APAData\GitHub\Pipeline\scipiper/DESCRIPTION' ... OK
* preparing 'scipiper':
* checking DESCRIPTION meta-information ... OK
* installing the package to build vignettes
* creating vignettes ... OK
Warning: 'inst/doc' files
    'shared_cache.Rmd', 'task_plans.Rmd', 'shared_cache.html',  'task_plans.html', 'shared_cache.R', 'task_plans.R'
  ignored as vignettes have been rebuilt.
  Run R CMD build with --no-build-vignettes to prevent rebuilding.
[...]

PR #91, opened 2018-12-20
A few more check fixes / attempts

aappling-usgs:
Following David's suggestion from the last PR, merging some upstream changes, and a small change to the Travis config that maybe will help. I've been unable to fix the vignette building issue locally; the vignettes build fine with `devtools::build_vignettes()` but break in the `R CMD check` context.


aappling-usgs:
Addresses #88

PR #95, opened 2019-02-05
add pitfalls vignette

aappling-usgs:
Looking for feedback on [vignettes/pitfalls.Rmd](https://github.com/USGS-R/scipiper/compare/master...aappling-usgs:pitfalls-vignette?expand=1#diff-f92259105d9c0fa3c06b5d4b0f49c0e1)


inst/doc/pitfalls.Rmd
```diff
@@ -0,0 +1,51 @@
...
+
+* <sym>Unnecessary re-runs of functions.</sym> `scipiper` and `remake` don't update their databases when you run a function outside of scipiper, even if the function creates the desired outputs. This may be frustrating if you've just run a long-running function outside of `scipiper` and are now attempting to use `scipiper` to build targets that depend on the function's output, because `scipiper` will insist on running that function again. <sol>Anytime you might want the output of a function to be considered complete, build the `scipiper` target rather than calling the function directly.</sol>
+
+* <sym>Common debugging methods don't work.</sym> If you build a `scipiper` target that runs function `A()` to create target `a`, setting RStudio checkpoints or `debug(A)` won't let you debug function `A` when you call `scmake('a')`. <sol>Temporarily insert a call to `browser()` within the function instead.</sol>
+
+* <sym>The GitHub repo is too big.</sym> You might be committing data files or large task remake files to the repo. <sol>In general, you can safely .gitignore all data files, the .remake/ folder, and (optionally) any task remake files.</sol>
```
jzwart: Creating `tmp/` directories for intermediate data processing steps and .gitignoring those folders is also common

inst/doc/pitfalls.Rmd
```diff
@@ -0,0 +1,51 @@
...
+
+* <sym>Teammates can't see updated objects.</sym> R objects can't be shared. <sol>To share objects, save them as .rds files and use the standard shared cache syntax.</sol>
+
+* <sym>Seemingly unnecessary builds of objects or files that a teammate already built.</sym> Those files may lack corresponding .ind files (and objects always lack .ind files), which are the key to avoiding rebuilds. With a shared cache approach, every data file or object not represented by an .ind file must be built locally by everybody who needs to build any target downstream of that file or object. <sol>For every file that takes a long time or special software to build: (1) use file rather than object targets, (2) define a corresponding .ind file and make downstream targets depend on the .ind file rather than the data file, (3) push the file to the shared cache, and (4) commit the .ind file to the repository.</sol>
```
jzwart:
```suggestion
* <sym>Seemingly unnecessary builds of objects or files that a teammate already built.</sym> Those files may lack corresponding `.ind` files (and objects always lack `.ind` files), which are the key to avoiding rebuilds. With a shared cache approach, every data file or object not represented by an `.ind` file must be built locally by everybody who needs to build any target downstream of that file or object. <sol>For every file that takes a long time or special software to build: (1) use file rather than object targets, (2) define a corresponding `.ind` file and make downstream targets depend on the `.ind` file rather than the data file, (3) push the data file to the shared cache, and (4) commit the `.ind` file to the repository.</sol>
```

inst/doc/pitfalls.Rmd
```diff
@@ -0,0 +1,51 @@
...
+
+* <sym>There are more .ind files than I want to manage.</sym> Not every data file (or object) needs to be in the shared cache, and anything not in the shared cache is a candidate for also not having an .ind file. If (a) you're comfortable requiring everybody to build a file or object locally, or (b) there are no declared downstream dependencies of the file/object, then you can skip creating an .ind file for it. For example, you might have a task remake file whose execution (e.g., with `loop_tasks()`) depends on `X` and `Y` and produces a single output file `Z`. As long as the main remake file says that `loop_tasks()` creates `Z.ind` (not `Z`), you can have just one person run the entire task remake file, and your team can share the output, without using any .ind files within those tasks. <sol>Don't create .ind files for targets that don't need them.</sol>
```
jzwart:
```suggestion
* <sym>There are more .ind files than I want to manage.</sym> Not every data file (or object) needs to be in the shared cache, and anything not in the shared cache is a candidate for also not having an `.ind` file. If (a) you're comfortable requiring everybody to build a file or object locally, or (b) there are no declared downstream dependencies of the file/object, then you can skip creating an `.ind` file for it. For example, you might have a task remake file whose execution (e.g., with `loop_tasks()`) depends on `X` and `Y` and produces a single output file `Z`. As long as the main remake file says that `loop_tasks()` creates `Z.ind` (not `Z`), you can have just one person run the entire task remake file, and your team can share the output, without using any `.ind` files within those tasks. <sol>Don't create `.ind` files for targets that don't need them.</sol>
```

inst/doc/pitfalls.Rmd
```diff
@@ -0,0 +1,51 @@
...
+
+## Team with multiple operating systems
```
jzwart: What about the graphic differences between Mac and Windows? This is a pretty specific pitfall to creating visualizations and I also can't remember if we got around the Windows poor graphic quality...

aappling-usgs: Good point, and no, we haven't solved it yet except by using scipiper so that somebody with a Mac can build the final versions. I'll add a bullet about this.

inst/doc/pitfalls.Rmd
```diff
@@ -0,0 +1,51 @@
...
+
+* <sym>The GitHub repo is too big.</sym> You might be committing data files or large task remake files to the repo. <sol>In general, you can safely .gitignore all data files, the .remake/ folder, and (optionally) any task remake files.</sol>
+
+## Use of shared cache
+
+* <sym>Teammates aren't getting updates about available intermediate data files.</sym> When sharing intermediate data files with a team using a shared cache, the status information about those files is contained in (1) a \*.ind file promising the existence of the data file in the cache, and (2) a build/status/\*.yml file telling `remake` how to decide whether the \*.ind file is up to date. <sol>Always `git commit` the \*.ind and build/status/\*.yml files corresponding to all shared data files.</sol>
+
+* <sym>Teammates can't see updated objects.</sym> R objects can't be shared. <sol>To share objects, save them as .rds files and use the standard shared cache syntax.</sol>
```
jread-usgs: is this the preferred format? (.rds)

aappling-usgs: Well...I prefer it to .RData, and besides .RData it's the other format that's great at handling any R object type you throw at it, but I didn't mean to restrict it here; .feather, .tsv, and anything else would be just fine, too. I'll change the language.

inst/doc/pitfalls.Rmd
lindsaycarr: You can only gitignore data files if you are using a shared cache approach, right? Or you expect it to build fairly quickly?

inst/doc/pitfalls.Rmd
lindsaycarr: I would like to suggest adding a section about building targets while you're testing code changes and not changing the shared cache until you are ready.

Here's my suggestion:

`<sym>`I am working on code that changes data in the shared cache but I'm not ready to commit those changes, and teammates are being forced to rebuild.`</sym>` There are many times where you are locally working on code that changes data on the shared cache. You might want to test that you can build the target as you are developing without committing these new changes, and subsequent `.ind` and build/status/\*.yml files. However, when you change the shared cache by executing the code you are still working on locally, other teammates who are building targets that depend on that one will be forced to rebuild the target because their `.ind` and build files show that they are outdated.`<sol>`Use `options(scipiper.dry_put = TRUE)` to prevent files built from automatically being pushed to the shared cache. Remember to revert to `options(scipiper.dry_put = FALSE)` when you are ready to push your code changes.`</sol>`

aappling-usgs: Oh, excellent idea! Thanks for the suggestion.

inst/doc/pitfalls.Rmd
aappling-usgs: No, I think you can and should generally gitignore data files for all projects, because GitHub (without GitLFS, anyway) isn't made to store big or binary files, and most data files are big and binary.

But if you don't use a shared cache, it's on you to store and back up the data files in some other way - could be as simple/risky as keeping the canonical copy in your local repo, or could involve doing the whole analysis on a shared drive, or running backups yourself, or manually copying files to and from some shared location. So I'll edit this text to add that caveat to the "safely" idea - data do need to be backed up somewhere, it's just that GitHub is usually a poor choice.

The one situation where I think it's a good idea to NOT .gitignore a data file is when it's (1) small, (2) ideally text-based so that the git diffs are meaningful to humans, and (3) possibly also not practical to store anywhere better. So we often make this exception for configuration files, small input data files that would be a bother to share via S3 or Drive, small input files when we're not doing any sharing elsewhere, and occasionally some files that summarize the status of the pipeline or the results as those change over time. We occasionally also make the exception for image files that we really want to see and store multiple versions of (e.g., you probably did this with gsplot?) - it's not that you _can't_ store binary files but that their storage often makes the repo too darn big.

inst/doc/pitfalls.Rmd
limnoliver: It might be worth explicitly stating that while you can reduce the number of indicator files with this behavior, if there are changes and _someone else tries run the task remake file_ - they'll have to rebuild everything locally. That is, the same person always needs to run the task remake file in order to not duplicate efforts. I think you imply this by saying "just one person". I ran into this issue with the lake temp cooperator files because I hadn't thought it through to this point!

PR #96, opened 2019-02-05
Shared cache vignette

jzwart:
Expanding upon Alison's shared cache vignette. 

I wasn't sure what the advantages are of using 2 vs. 3 targets per data file so I left that blank for now. We'll discuss during the data pipelines meeting.


vignettes/shared_cache.Rmd
```diff
@@ -8,45 +12,154 @@ vignette: >
...
-* Intermediate and final products can be immediately visible to anyone who has access to the shared cache, whether they are contributing to the analysis or simply inspecting/using the output.
+A **s**hared **c**ache (**sc** of `scmake`) is a cloud data storage location where raw, intermediate, and/or final data products from an analysis project are contributed to and accessible by multiple analysts. Not all scipiper projects will use a shared cache.
+
+<data>**Data files**</data> only _need_ to be local when the analyst is computing with the <data>**data file**</data>. <ind>**Indicator files**</ind> (`.ind`) represent the remote shared cache among project participants. This allows analyst #1 to compute part of the project, upload output in the project's shared cache, and analyst #2 can use the output from analyst #1 without redoing the computing performed by analyst #1 
+
+Workflow dependencies are connected via the <ind>**indicator files**</ind>. Recipes (e.g. R scripts) used to create the <ind>**indicator files**</ind> pull down the <data>**data files**</data> from the shared cache using `scipiper` functions such as `gd_get()`.
```
aappling-usgs: I'd make a distinction between the creation of an indicator file, which usually involves pushing a data file _to_ the shared cache, from the use of an indicator file, which usually involves pulling a data file _from_ the shared cache (if it's not already available locally).

vignettes/shared_cache.Rmd
```diff
@@ -8,45 +12,154 @@ vignette: >
...
+
+<data>**Data files**</data> only _need_ to be local when the analyst is computing with the <data>**data file**</data>. <ind>**Indicator files**</ind> (`.ind`) represent the remote shared cache among project participants. This allows analyst #1 to compute part of the project, upload output in the project's shared cache, and analyst #2 can use the output from analyst #1 without redoing the computing performed by analyst #1 
+
+Workflow dependencies are connected via the <ind>**indicator files**</ind>. Recipes (e.g. R scripts) used to create the <ind>**indicator files**</ind> pull down the <data>**data files**</data> from the shared cache using `scipiper` functions such as `gd_get()`. 
+
+<build>**Build files**</build> tell remake if the <ind>**indicator files**</ind> (and other <target>**targets**</target> that are <data>**data files**</data>) and their dependencies are up-to-date. <target>**Targets**</target> that are not saved as <ind>**indicator**</ind> or <data>**data**</data> files (e.g. R objects) will not have a <build>**build status file**</build> and will be excuted every time it is called.
```
aappling-usgs: We actually don't create build/status/*.yml files for regular old data files, only for indicator files.

vignettes/shared_cache.Rmd
```diff
@@ -8,45 +12,154 @@ vignette: >
...
+
+targets:
+
+  <target>1_data:</target>
+     depends:
+      - <ind>1_data/out/compiled_data.rds.ind</ind>
+      - <ind>1_data/out/selected_sites.rds.ind</ind>
+
+  <ind>1_data/out/compiled_data.rds.ind:</ind>
+    command: gather_stream_data(
```
aappling-usgs: Could rename to something like `gather_and_share_stream_data` to clarify that pushing to GD happens here.

aappling-usgs: And actually, would you be up for also creating code snippets for `gather_[and_share]_stream_data` and `select_sites` that include the `gd_put` and `gd_get` calls? They're a useful part of the overall SC pattern, I think.

vignettes/shared_cache.Rmd
```diff
@@ -8,45 +12,154 @@ vignette: >
...
-* Though we've done much to ensure this doesn't happen, it's conceivable that metadata will become corrupt relative to the data. Some monitoring and very occasional full rebuilding is recommended when practical.
-* Old files no longer referenced by the code can accumulate on the shared cache unless manually deleted. Though these will not interfere with ongoing analysis, they can take up storage space unnecessarily.
 
 ### Guidelines
 
-Projects using a shared cache should follow these guidelines:
+####Projects using a shared cache should follow these guidelines:
+
+* Use <ind>**indicator files**</ind> (usually an `.ind` suffix) to represent most or all of the chain of connected <target>**targets**</target> within your remake files. Each `.ind` file should be one of two products of a recipe (e.g. R script), where the other product is the creation of a <data>**data file**</data>, either locally and/or in the shared cache.
```
aappling-usgs: I'd define a recipe as an R function call rather than an R script. We sometimes write multiple functions in the same script file, and sometimes multiple recipes use the same function but each call it differently.

aappling-usgs: Consider pointing out that the .ind file is the only product of the recipe that is declared to remake; the data file is hidden from / not announced to remake, and the purpose of the .ind file is to indirectly represent the data file to remake (where the indirectness is what enables compatibility between remake and the shared cache).

vignettes/shared_cache.Rmd
```diff
@@ -8,45 +12,154 @@ vignette: >
...
+
+* Generally avoid using R objects as `remake/scmake` <target>**targets**</target>...but if you must, usually for convenience or conciseness of the workflow plan, recognize that R objects must be built by every analyst. So if a target takes a non-trivial length of time to build, or if it depends on large volumes of <data>**data**</data> as input, that target should usually be a file rather than an R object.
 
-* Use indicator files (usually an `.ind` suffix) to represent most or all of the chain of connected targets within your remake files. Each .ind file should be one of two products of a recipe, where the other product is the creation of a data file, either locally and/or in the shared cache.
+* `git commit` all <ind>**indicator files**</ind> and <build>**build files**</build> (with occasional exception of <ind>**indicator files**</ind> within a task plan; those require additional thought). `git ignore` all <data>**data files**</data> *except* those that are  necessary to pipeline basic functions (e.g. <data>**gd_config.yml**</data>).
```
aappling-usgs: I think the choice of which data files to commit is not so much about necessity but rather about whether they are amenable to storage in git/GitHub - relatively small, text-based, typically hand-curated, and having their canonical copy within the git repo rather than in the shared cache or another cloud location. For example, another data file we might commit could be a short .tsv file that maps lake IDs to the filenames of the corresponding collaborator-provided data files.

While for most data files I see git committing as optional, enabled but not required by the above criteria. In contrast, gd_config.yml really _should_ be git committed because it informs the basic pipeline function of data exchange with Google Drive...but I'd argue that gd_config.yml is only borderline a "data file" anyway, because its "data" are configuration parameters.

vignettes/shared_cache.Rmd
```diff
@@ -1,5 +1,9 @@
+## Shared Cache {.tabset .tabset-fade .tabset-pills}
+
+Alison Appling, July 31, 2018;  
+Jake Zwart, February 4, 2019
+
 ---
-title: "Shared cache"
 author: "Alison Appling"
```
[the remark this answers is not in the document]
jzwart: Yeah I think it's the `{.tabset}` that's screwing that up, I'll remove the extra lines

vignettes/shared_cache.Rmd
```diff
@@ -8,45 +12,154 @@ vignette: >
...
 
-Advantages of a shared cache:
-* Not every analyst needs to build every target, saving on total processing time.
-* Targets that can only be built on specific operating systems (e.g., Mac) or in specific computing environments (e.g., a cluster) can still be accessible to all analysts for further analysis.
-* Intermediate and final products can be immediately visible to anyone who has access to the shared cache, whether they are contributing to the analysis or simply inspecting/using the output.
+A **s**hared **c**ache (**sc** of `scmake`) is a cloud data storage location where raw, intermediate, and/or final data products from an analysis project are contributed to and accessible by multiple analysts. Not all scipiper projects will use a shared cache.
+
+<data>**Data files**</data> only _need_ to be local when the analyst is computing with the <data>**data file**</data>. <ind>**Indicator files**</ind> (`.ind`) represent the remote shared cache among project participants. This allows analyst #1 to compute part of the project, upload output in the project's shared cache, and analyst #2 can use the output from analyst #1 without redoing the computing performed by analyst #1
```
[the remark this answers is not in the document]
jzwart: good idea, I'll add those details about intermediate data files / inputs

vignettes/shared_cache.Rmd
jzwart: Ah good distinction, I'll add that

vignettes/shared_cache.Rmd
jzwart: Oh oops I'll remove

vignettes/shared_cache.Rmd
jzwart: Yes, I'll add some code snippets of those functions that get called in the targets

vignettes/shared_cache.Rmd
jzwart: Thanks for the clarification on that, I'll make that point.

vignettes/shared_cache.Rmd
jzwart: I'll expand on the data file explanation for what can / should get committed. Thanks!

vignettes/shared_cache.Rmd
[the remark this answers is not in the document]
aappling-usgs: I wouldn't go as far as "screwing that up" - I really like how the tabset theme looks and works!

inst/doc/shared_cache.Rmd
```diff
@@ -1,52 +1,191 @@
...
-## Setup
 
-```{r libs, message=FALSE}
-library(scipiper)
-library(dplyr)
-```
+### Terms and concept
 
-## Shared cache
+####Terms
```
lindsaycarr: Should we include a definition of the yaml file here?

inst/doc/shared_cache.Rmd
```diff
@@ -1,52 +1,191 @@
...
-* Old files no longer referenced by the code can accumulate on the shared cache unless manually deleted. Though these will not interfere with ongoing analysis, they can take up storage space unnecessarily.
 
 ### Guidelines
 
-Projects using a shared cache should follow these guidelines:
+####Projects using a shared cache should follow these guidelines:
+
+* Use <ind>**indicator files**</ind> (usually an `.ind` suffix) to represent most or all of the chain of connected <target>**targets**</target> within your remake files. Each `.ind` file should be one of two products of a recipe (e.g. R function call), where the other product is the creation of a <data>**data file**</data>, either locally and/or in the shared cache. The <ind>**indicator files**</ind> are the only products that are declared to `remake`, while the <data>**data files**</data> remain hidden. This allows the <ind>**indicator files**</ind> to indirectly represent the <data>**data files**</data> to `remake` thereby enabling compatibility between `remake` and the shared cache.
```
lindsaycarr: I would suggest adding an example of what doesn't need to have an `ind` file. Such as things that are quick to build (like reading info from a yaml file for state names or something), maybe targets that require < 5 seconds of build time.

inst/doc/shared_cache.Rmd
```diff
@@ -1,52 +1,191 @@
...
+
+* Always build <target>**targets**</target> using `scipiper::scmake()` rather than `remake::make()`. Though the functions are outwardly very similar, `scmake()` maintains an extra layer of metadata that allows multiple users to share a single project <build>**build status**</build> (e.g., "file x.rds.ind is up to date; file y.rds.ind is out of date"). In a shared-cache project, you should not even need to load the `remake` package directly.
+
+* Generally avoid using R objects as `remake/scmake` <target>**targets**</target>...but if you must, usually for convenience or conciseness of the workflow plan, recognize that R objects must be built by every analyst. So if a target takes a non-trivial length of time to build, or if it depends on large volumes of <data>**data**</data> as input, that target should usually be a file rather than an R object.
```
lindsaycarr: Why should you generally avoid this? I have a lot of code that does this - particularly when reading from config files.

inst/doc/shared_cache.Rmd
```diff
@@ -1,52 +1,191 @@
...
+
+* `git commit` all <ind>**indicator files**</ind> and <build>**build files**</build> (with occasional exception of <ind>**indicator files**</ind> within a task plan; those require additional thought). `git ignore` all <data>**data files**</data> unless they are small enough to store in git/GitHub, such as small, text-based, and typically hand-curated <data>**data files**</data> (e.g. a <data>**data file**</data> that matches NHD lake ID's to collaborator-provided <data>**data files**</data>). <data>**Data files**</data> that are  necessary to pipeline basic functions should also be committed (e.g. <data>**gd_config.yml**</data>). 
+
+* To force a rebuild, either use the `force=TRUE` argument to `scmake()` or use `scdel()` to delete <ind>**indicator files**</ind>. There's seldom any benefit to deleting <data>**data files**</data> (by any method); usually deleting the <ind>**indicator files**</ind> is plenty. When deleting <ind>**indicator files**</ind>, `force=TRUE` or `scdel()` are preferable to directly deleting the `.ind` files because if only the `.ind` files are deleted, the `scipiper` database may fail to update properly when the `.ind` files are rebuilt.
```
lindsaycarr: Should we mention anything about touching (or most likely not touching) the `remake/build` files?

inst/doc/shared_cache.Rmd
```diff
@@ -1,52 +1,191 @@
...
+  <data>1_data/out/nitrate_data_pull.rds:</data>
+    command: gd_get(<ind>'1_data/out/nitrate_data_pull.rds.ind'</ind>, config_file = <data>'lib/cfg/gd_config.yml'</data>)
 
-* Use indicator files (usually an `.ind` suffix) to represent most or all of the chain of connected targets within your remake files. Each .ind file should be one of two products of a recipe, where the other product is the creation of a data file, either locally and/or in the shared cache.
+</pre>
 
-* Always build targets using `scipiper::scmake()` rather than `remake::make()`. Though the functions are outwardly very similar, `scmake()` maintains an extra layer of metadata that allows multiple users to share a single project status (e.g., "file x.rds.ind is up to date; file y.rds.ind is out of date"). In a shared-cache project, you should not even need to load the `remake` package directly.
+#####2 <target>**target**</target> method: 
+1. push a <data>**data file**</data> and create an <ind>**indicator file**</ind> 
+2. retrive the <data>**data file**</data>
```
lindsaycarr: ~retrive~ retrieve

inst/doc/shared_cache.Rmd
```diff
@@ -1,52 +1,191 @@
...
+
+* To force a rebuild, either use the `force=TRUE` argument to `scmake()` or use `scdel()` to delete <ind>**indicator files**</ind>. There's seldom any benefit to deleting <data>**data files**</data> (by any method); usually deleting the <ind>**indicator files**</ind> is plenty. When deleting <ind>**indicator files**</ind>, `force=TRUE` or `scdel()` are preferable to directly deleting the `.ind` files because if only the `.ind` files are deleted, the `scipiper` database may fail to update properly when the `.ind` files are rebuilt.
+
+####How many <target>targets</target> should I use per <data>data file</data>?
+
+#####3 <target>**target**</target> method: 
+1. create a <data>**data file**</data> 
+2. push a <data>**data file**</data> and create an <ind>**indicator file**</ind> 
+3. retrive the <data>**data file**</data>
```
lindsaycarr: ~retrive~ retrieve

inst/doc/shared_cache.Rmd
ldecicco-USGS: Yeah, is this the consensus? And if so, I'd also vote to have some explanation why to avoid objects. That was one of my main reasons for preferring `remake` to `make`.

aappling-usgs: But surely you tend to prefer objects over files specifically when they take a trivial length of time to build or depend on less-than-large volumes of data as input (config files being a good example)? Or according to some other considerations?

@wdwatkins is scheduled to do some serious thinking and writing about this for our march 26 pipelines meeting:` David - March 26 - Doc: When to use R objects vs data files vs indicator+data files` - let's make sure we get any related thoughts over to him for that.

ldecicco-USGS: @aappling-usgs ...I'm not sure if your question is directed at me or Jake...so if me: My preference is to default to objects unless there is a reason to have to deal with files, and I guess if you've committed some target to following the shared cache approach, that would be a reason for a file (?). The sentence made it seem like all scipiper targets should avoid objects.  Maybe all we need to do is "Generally avoid using R objects as shared-cache targets"? 

I'm hoping David's vignette will tease out some benchmarks for a file (file size, sharing with non-R users, shared cache that needs to be pushed up to AWS/GoogleDrive/etc...and *that* probably depends on how long you are willing to wait for computations).

aappling-usgs: (Yes, my comment was directed at Laura and Lindsay's comments.)

Good point that the guidelines for object vs file should be quite different for shared-cache vs non-shared-cache projects (and could be more like non-shared-cache guidelines for specific task tables even within generally-shared-cache projects). I'm also intrigued by the challenge to provide specific benchmarks.

Note that this particular vignette and PR are for a vignette specifically about shared caches.

inst/doc/shared_cache.Rmd
wdwatkins: Yeah, when building is likely to be faster than downloading

inst/doc/shared_cache.Rmd
jzwart: Yeah I think that would be helpful. I can add a short description

inst/doc/shared_cache.Rmd
jzwart: good point, I will add that

inst/doc/shared_cache.Rmd
jzwart: I'll specify in the paragraph that it's focusing on shared cache and also that there may be thresholds / guidelines to file size, etc... for when to do R object vs. file (this can get updated after David's pipelines meeting)

PR #103, opened 2019-02-28
switch to pkgdown

aappling-usgs:
This PR successfully builds the pkgdown site for all documented functions and the shared_cache.Rmd vignette, but the current task_plans.Rmd vignette doesn't build (`Reading 'vignettes/task_plans.Rmd'...Error in file(con, "r") : cannot open the connection`). Merging so that Lindsay can update her fork (with the new task_plans vignette) and Laura can help debug from there.


aappling-usgs:
Whoops, might have been a false start on Travis because I did some accidental git gymnastics locally. Going to close and try again in case that was the issue.

PR #104, opened 2019-02-28
switch to pkgdown

aappling-usgs:
This PR successfully builds the pkgdown site for all documented functions and the shared_cache.Rmd vignette, but the current task_plans.Rmd vignette doesn't build (`Reading 'vignettes/task_plans.Rmd'...Error in file(con, "r") : cannot open the connection`). Merging so that Lindsay can update her fork (with the new task_plans vignette) and Laura can help debug from there.

(retry of #103)


aappling-usgs:
I'm going to cross fingers and merge despite the travis errors, because those seem to be (i hope!) just a wonky github thing. I will fix if the test on master doesn't work.

PR #105, opened 2019-02-28
try again for travis

aappling-usgs:
Here trying for a travis run where inst/doc is already available (vignettes are pre-built).


aappling-usgs:
I think I messed this up.

PR #108, opened 2019-03-07
adding task_plan back in to be reviewed

lindsaycarr:
The HTML has already been committed if you'd like to view it in the right formatting to review: https://usgs-r.github.io/scipiper/task_plans.html.


vignettes/task_plans.Rmd
```diff
@@ -0,0 +1,353 @@
...
+1. baked-in retries during each execution of `loop_tasks`
+2. quicker restarts if `loop_tasks` gets interrupted partway through (e.g., if your computer runs out of battery or crashes)
+3. the ability to easily run subsets of tasks and/or steps while you're developing the analysis (see below)
+
+To use the baked-in retries effectively, simply ensure that the functions called in your task-steps throw errors (typically with `stop()`) when things go wrong. No need to build retries into the functions themselves because `loop_tasks` will take care of that for you. And if your function isn't the sort that would benefit from retries - e.g., HTTP requests need retries, but a deterministic algorithm should just give up if it breaks - you can set `num_tries=1` in `loop_tasks`.
+
+
+### Forcing rebuilds
+
+#### Forcing rebuild for one task-step
```
jzwart:
```suggestion
#### Forcing rebuild for one task step
```

vignettes/task_plans.Rmd
```diff
@@ -0,0 +1,353 @@
...
+
+Similar to the one-target example, above, if we run the same `loop_tasks` command twice, nothing will get rebuilt the second time. Instead, all items will say `OK`:
+
+```{r try_rebuild_all}
+loop_tasks(
+  task_plan=task_plan_1,
+  task_makefile='task_plan_1.yml')
+```
+
+One way to force a complete rebuild (not generally a good use of time, but sometimes necessary) is to use `force=TRUE` argument to `loop_tasks`. This will automatically use `scdel` with all `target_names` in `task_plan_1.yml`. Now when we run the entire loop, everything gets rebuilt:
```
jzwart: Could you give a quick example of when this might be necessary?

vignettes/task_plans.Rmd
```diff
@@ -0,0 +1,353 @@
...
+
+<head>
+<style>
+data {color: #CA3542;}
+ind {color: #FEBC38;}
+build {color: #697F98;} 
+target {color: #37AFA9}
+yml {color: #56B4E9}
+</style>
+</head>
```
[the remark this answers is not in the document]
lindsaycarr: Good catch; I will delete. Originally kept them there so that if I wanted to use colors, I would remember how.

vignettes/task_plans.Rmd
```diff
@@ -0,0 +1,353 @@
...
+
+```{r makefile, results='markup'}
+create_task_makefile(
+  task_plan=task_plan_1,
+  makefile='task_plan_1.yml',
+  include=c(),
+  packages=c('ggplot2'),
+  sources=c('script.R'),
+  file_extensions=c('ind'),
+  ind_complete=TRUE)
```
aappling-usgs: I think we'll need to drop `ind_complete` from the arg list above when we merge the `task_combiners` branch into `master`.

aappling-usgs: We could drop it right now, in anticipation of that change; the only awkwardness would be that the above command would yield `message('ind_complete=NA is deprecated; use TRUE or FALSE')` which we'd have to ignore for another few weeks to months. I think dropping it right now would be better than not dropping it and forgetting until the vignette breaks.

vignettes/task_plans.Rmd
```diff
@@ -0,0 +1,353 @@
...
+loop_tasks(
+  task_plan=task_plan_1,
+  task_makefile='task_plan_1.yml')
+```
+
+The benefits of using `loop_tasks` to run all the remake targets in task_plan_1.yml are:
+
+1. baked-in retries during each execution of `loop_tasks`
+2. quicker restarts if `loop_tasks` gets interrupted partway through (e.g., if your computer runs out of battery or crashes)
+3. the ability to easily run subsets of tasks and/or steps while you're developing the analysis (see below)
```
aappling-usgs: Change "see below" to "see next section" now that we're using tabbed sections

vignettes/task_plans.Rmd
```diff
@@ -0,0 +1,353 @@
...
+
+To use the baked-in retries effectively, simply ensure that the functions called in your task-steps throw errors (typically with `stop()`) when things go wrong. No need to build retries into the functions themselves because `loop_tasks` will take care of that for you. And if your function isn't the sort that would benefit from retries - e.g., HTTP requests need retries, but a deterministic algorithm should just give up if it breaks - you can set `num_tries=1` in `loop_tasks`.
+
```
aappling-usgs: It might be good to add a sentence or paragraph here about how we've learned that `loop_tasks` can be waaaay slower than a simple `scmake` for large task makefiles. I wish it just weren't slower, but I don't think we're going to fix that anytime soon, and in the meantime, it pays to know that you have both options. The  command for running this task plan with `scmake` would be
```r
scmake('task_plan_1', 'task_plan_1.yml')
```
and a remake recipe for it in the parent makefile might be
```yml
  states/log/task_plan_1.ind:
    command: scmake('task_plan_1', 'task_plan_1.yml', force=TRUE)
```
where the `force=TRUE` within the command would mean that even though the file `states/log/task_plan_1.ind` only gets built once on its own, you could call `scmake('states/log/task_plan_1.ind', force=TRUE)` at the prompt and because the recipe command also has force=TRUE, the indicator file would actually get rebuilt. (This still doesn't rebuild the individual task-steps, but it does avoid one sometimes-surprising behavior (failure to rebuild when you call force=TRUE at the prompt but not in the parent makefile).

aappling-usgs: ...And maybe that last idea, the addition of `force=TRUE`, would be good to save for the next section.

vignettes/task_plans.Rmd
aappling-usgs: It could be necessary when your pipeline doesn't capture all the dependencies (not ideal, but also not uncommon). For example, maybe your first task_step is to pull data from NWIS - you'd want force=TRUE when you were ready to pull  a new round of updated data from NWIS, and force=TRUE would be necessary because nothing in your local files exists to tell you that the NWIS data have been changing online.

vignettes/task_plans.Rmd
lindsaycarr: Seems almost as simple as deleting, but there is a section that talks a lot about this option. See the `### Indicator files options` section. I think we would need to change more than just that argument when task_combiners gets merged.

aappling-usgs: OK, waiting sounds like a good plan :+1:

aappling-usgs:
Oh, Travis + R-devel. =(

I wonder if the new error could be because of the change in random seeds that Laura mentioned recently?
```
  ── 1. Failure: parallel loop_tasks completes while handling errors (@test-loop_t
  length(grep("Starting loop attempt", output)) is not strictly more than 1. Difference: 0
```

aappling-usgs:
Huzzah, useful vignette AND passing travis!!

PR #110, opened 2019-03-26
target types vignette

wdwatkins:


vignettes/target_types.Rmd
```diff
@@ -0,0 +1,49 @@
...
+
+`scipiper` targets can be indicator files with a [shared cache](shared_cache.html), local data files with or without local indicator files, or simply R objects.  This vignette provides some guidelines for when to use these different patterns.
+
+## Objectives
+There are two main goals to keep in mind when deciding what style of target to use:
+ 
+ - **Minimize time spend unnecessarily rebuilding targets** --- i.e., those that have already been built locally or by collaborators
+ - **Keeping the pipeline simple** --- caching can save lots of build time, but also adds to the code that has to be maintained, can clutter the repo with build and indicator files, etc.
+ 
+ These two can sometimes be in tension, but usually it will be clear which is more important for a given case.
```
aappling-usgs: Nice distillation.

vignettes/target_types.Rmd
```diff
@@ -0,0 +1,49 @@
...
+ 
+## Key factors to consider
+
+ - **Time to create an object** vs **object size**: This is the biggest thing to consider --- is it faster to upload a file to the shared cache and have others download it, or just have others rebuild it themselves?  This will somewhat depend on your internet connection.  Don't forget about upload time as well.  Uploading is generally slower than downloading, especially for remote workers with residential internet connections.
```
aappling-usgs: I'm skeptical that our residential internet connections are noticeably slower than the Middleton connection... :stuck_out_tongue_winking_eye:

vignettes/target_types.Rmd
```diff
@@ -0,0 +1,49 @@
...
+
+## Objectives
+There are two main goals to keep in mind when deciding what style of target to use:
+ 
+ - **Minimize time spend unnecessarily rebuilding targets** --- i.e., those that have already been built locally or by collaborators
+ - **Keeping the pipeline simple** --- caching can save lots of build time, but also adds to the code that has to be maintained, can clutter the repo with build and indicator files, etc.
+ 
+ These two can sometimes be in tension, but usually it will be clear which is more important for a given case.
+ 
+## Key factors to consider
```
aappling-usgs: These factors apply specifically to the question of whether something should be in the shared cache or not. There's also the question of whether something local-only should be an object or a file. Ideally this vignette would discuss both questions because they so often come up together.

vignettes/target_types.Rmd
```diff
@@ -0,0 +1,49 @@
...
+ 
+ - **How often will a target be updated?**:  Saving a target to disk may be less useful if it is likely to need to be rebuilt before the file on disk is ever used again.
+ 
+ - **Collaborators**: Are all collaborators equipped to rebuild every step in the pipeline, if needed?  If a particular step needs to be done by certain people, say a processing-intensive munging step early in the pipeline, than this file will almost certainly need to be in the remote cache.
+ 
+## Target choices
+
+**Shared cache data/indicator file**: The target will be only be rebuilt if its dependencies have changed, and the indicator file has not been updated to trigger a download from the remote cache. For longer-running operations that are worthwhile to cache remotely.  See the [shared caching vignette](shared_cache.html) for different implementation options.
```
aappling-usgs: Clarification: even if the indicator file has been updated, the data file target will only be rebuilt (i.e., downloaded) if the data file is explicitly required and requested for some other operation. This is because .yml files designed for a shared cache [should] only tell remake about dependencies among .ind files, and they don't/shouldn't tell remake about how data files relate to one another. Which means that if I only need some data file far downstream of some indicator file, I should never actually need to download that file from the remote cache at all. In those cases, the uploading and storage are issues but the downloading cost is probably pretty low for the team as a whole.

Agreed on the primary use case (use for longer-running operations that are worthwhile to cache remotely).

However, I think there's also a secondary use case for having an indicator file. If you don't use an indicator file, then everybody needs to build the data file (or object) not just to be able to use that file/object directly, but also to be able to acquire or build _any_ file _anywhere_ downstream of the data file/object. So git-committed indicator files can be useful when you're coding a whole bunch of data targets that nobody else is going to need to use, and all they want is the output of your chain of analyses. Even if each of those data targets is relatively quick to build on its own, the whole set of targets can take a while. In that situation, using indicator files would save everybody else from having to build all of the intermediate data files, allowing them to jump straight to the files they actually need.

For the secondary use case, I suspect there's a way to include indicator files without actually pushing the data files to a shared cache, and I think we've maybe sorta done this with task plan outputs in the past, but the right method is still pretty fuzzy to me; it'd be great to get a recommendation written out.

vignettes/target_types.Rmd
```diff
@@ -0,0 +1,49 @@
...
+ These two can sometimes be in tension, but usually it will be clear which is more important for a given case.
+ 
+## Key factors to consider
+
+ - **Time to create an object** vs **object size**: This is the biggest thing to consider --- is it faster to upload a file to the shared cache and have others download it, or just have others rebuild it themselves?  This will somewhat depend on your internet connection.  Don't forget about upload time as well.  Uploading is generally slower than downloading, especially for remote workers with residential internet connections.   
+ 
+ - **How often will a target be updated?**:  Saving a target to disk may be less useful if it is likely to need to be rebuilt before the file on disk is ever used again.
+ 
+ - **Collaborators**: Are all collaborators equipped to rebuild every step in the pipeline, if needed?  If a particular step needs to be done by certain people, say a processing-intensive munging step early in the pipeline, than this file will almost certainly need to be in the remote cache.
+
```
aappling-usgs: Maybe add one bullet here about what I discuss in the following comment: Will all collaborators need to have this file locally, or can they skip having a local copy entirely, and if so, does it save a non-trivial amount of time and/or space for them to skip having a local copy?

vignettes/target_types.Rmd
```diff
@@ -0,0 +1,49 @@
...
+ 
+ - **How often will a target be updated?**:  Saving a target to disk may be less useful if it is likely to need to be rebuilt before the file on disk is ever used again.
+ 
+ - **Collaborators**: Are all collaborators equipped to rebuild every step in the pipeline, if needed?  If a particular step needs to be done by certain people, say a processing-intensive munging step early in the pipeline, than this file will almost certainly need to be in the remote cache.
+ 
+## Target choices
+
+**Shared cache data/indicator file**: The target will be only be rebuilt if its dependencies have changed, and the indicator file has not been updated to trigger a download from the remote cache. For longer-running operations that are worthwhile to cache remotely.  See the [shared caching vignette](shared_cache.html) for different implementation options.
+
+**Local data file only**:  The target will be rebuild if the local data file does not exist or depedencies have changed.  For cases where a target takes a relatively long time to build, but creates a massive file that is more cumbersome to upload/download than rebuild.
```
aappling-usgs: Or if the target is quick to build, right?

aappling-usgs:
```suggestion
**Local data file only**:  The target will be rebuilt if the local data file does not exist or depedencies have changed.  For cases where a target takes a relatively long time to build, but creates a massive file that is more cumbersome to upload/download than rebuild.  
```

vignettes/target_types.Rmd
```diff
@@ -0,0 +1,49 @@
...
+ 
+## Target choices
+
+**Shared cache data/indicator file**: The target will be only be rebuilt if its dependencies have changed, and the indicator file has not been updated to trigger a download from the remote cache. For longer-running operations that are worthwhile to cache remotely.  See the [shared caching vignette](shared_cache.html) for different implementation options.
+
+**Local data file only**:  The target will be rebuild if the local data file does not exist or depedencies have changed.  For cases where a target takes a relatively long time to build, but creates a massive file that is more cumbersome to upload/download than rebuild.  
+
+**Local data file with local indicator**: The target will be rebuilt if the indicator file is missing or dependencies have changed.  This is typically used to speed up `scipiper` by reducing redundant hashing of large files, or to represent multiple files, allowing a "one-to-many" target.
+
+**R object**: The target will be rebuilt every time.  For quick-building targets that aren't worth the time to write to disk, much less upload, e.g. most plots.
```
aappling-usgs: It's still only rebuilt every time it's missing or out of date on the local computer. We still get the standard remake benefits; it's just that every team member needs to build it locally once. This is another place where I think this document would be stronger if we could find a way to talk both about the question of shared cache vs unshared and the question of local file vs object.

Interesting that you single out plots. I think first of small data objects (e.g., configuration snippets, tiny data.frames, character vectors) that will be applied or combined to make something more broadly useful in subsequent steps. Or data that are changing really frequently as we write more code, going to your point about "How often will a target be updated?"

I actually think of plots as often being a really good thing to push to a shared cache, because often I want collaborators to be able to see and discuss the plots with me. Unless the collaborators are actively keeping up to speed with building a pipeline locally (definitely not the case for our non-R collaborators, and sometimes not the case even for R users), it's a lot quicker for us all to browse to a Drive folder than for everybody to build the plots locally.

vignettes/target_types.Rmd
```diff
@@ -0,0 +1,49 @@
...
+ 
+## Target choices
+
+**Shared cache data/indicator file**: The target will be only be rebuilt if its dependencies have changed, and the indicator file has not been updated to trigger a download from the remote cache. For longer-running operations that are worthwhile to cache remotely.  See the [shared caching vignette](shared_cache.html) for different implementation options.
+
+**Local data file only**:  The target will be rebuild if the local data file does not exist or depedencies have changed.  For cases where a target takes a relatively long time to build, but creates a massive file that is more cumbersome to upload/download than rebuild.  
+
+**Local data file with local indicator**: The target will be rebuilt if the indicator file is missing or dependencies have changed.  This is typically used to speed up `scipiper` by reducing redundant hashing of large files, or to represent multiple files, allowing a "one-to-many" target.
```
aappling-usgs: This bullet may need elaboration for newcomers and posterity - can you make these examples more concrete and detailed?

vignettes/target_types.Rmd
aappling-usgs: ...and now that I've read and thought through the full vignette, I'm wondering, which of these considerations applies when choosing between a local data file and a local R object? Is there a third consideration that needs to be added?

vignettes/target_types.Rmd
```diff
@@ -0,0 +1,49 @@
...
+  %\VignetteIndexEntry{Target types}
+  %\VignetteEngine{knitr::rmarkdown}
+  \usepackage[utf8]{inputenc}
+---
+
+```{r setup, include=FALSE}
+knitr::opts_chunk$set(echo = TRUE)
+```
+
+`scipiper` targets can be indicator files with a [shared cache](shared_cache.html), local data files with or without local indicator files, or simply R objects.  This vignette provides some guidelines for when to use these different patterns.
```
aappling-usgs: The shared_cache vignette covers it in more detail, but I think it'd be useful in this vignette to remind readers of how indicator file + data file pairs get used and why they work. The most important point, I think, is that when using indicator/data pairs, the workflow dependencies are connected via indicator files and not by data files, which means that remake ensures the local up-to-dateness of the indicator files, but corresponding data files don't necessarily need to be local and up to date, saving team members from unnecessary downloads or rebuilds of data files they're not using directly.

aappling-usgs: (That idea could surely be rephrased into shorter, more readable sentences than what I just wrote.)

vignettes/target_types.Rmd
aappling-usgs: Note that downloading is sometimes optional for many teammates, such that the file may only need to be present on one person's computer and the shared cache.

aappling-usgs: The title to this bullet is about time vs size, but the bullet text focuses on size only as it relates to internet transfer speed. That's an important consideration, but if we're not also concerned about storing large files on a shared cache because of the storage space/cost itself, maybe the title should call out transfer time rather than object size.

vignettes/target_types.Rmd
```diff
@@ -0,0 +1,49 @@
...
+ - **Minimize time spend unnecessarily rebuilding targets** --- i.e., those that have already been built locally or by collaborators
+ - **Keeping the pipeline simple** --- caching can save lots of build time, but also adds to the code that has to be maintained, can clutter the repo with build and indicator files, etc.
+ 
+ These two can sometimes be in tension, but usually it will be clear which is more important for a given case.
+ 
+## Key factors to consider
+
+ - **Time to create an object** vs **object size**: This is the biggest thing to consider --- is it faster to upload a file to the shared cache and have others download it, or just have others rebuild it themselves?  This will somewhat depend on your internet connection.  Don't forget about upload time as well.  Uploading is generally slower than downloading, especially for remote workers with residential internet connections.   
+ 
+ - **How often will a target be updated?**:  Saving a target to disk may be less useful if it is likely to need to be rebuilt before the file on disk is ever used again.
```
aappling-usgs: Great point. Can you give an example? The parsed version of viz_config.yml in vizstorm projects comes to mind for me; are there others that have been very clear use cases for you?

Also, every target gets saved to disk, even the "R objects", in remake land. So I think the thing we're choosing is not between saving and not saving a target to disk, between saving to a human-discoverable file (which can be useful for manual inspection...sometimes...) vs keeping the data stored in a hidden repository (and thus keeping our main repo less cluttered).

vignettes/target_types.Rmd
wdwatkins: Re: speeds -- my downstream is definitely faster than the office, but my upload may actually be worse.  Residential internet usually has pretty heavily throttled upload speeds, whereas business connections are more equal upstream/downstream.

vignettes/target_types.Rmd
wdwatkins: Mm good point — my point of view is mainly from building the analytics log processing pipelines, where sharing plots, and the plots need to be in memory so they can be passed to bookdown

wdwatkins: And I did not know that all objects get written to disk behind the scenes! I guess for local-only targets, it just comes down to whether you want easy access without going through remake.

vignettes/target_types.Rmd
wdwatkins: Do any of the recent features make the one-to-many application obsolete?

vignettes/target_types.Rmd
wdwatkins: Added **Do you need easy access to the underlying file**

vignettes/target_types.Rmd
wdwatkins: Ok — in that case, I think this falls into the new consideration of whether you want to look at the file or not, and this one can be removed.

vignettes/target_types.Rmd
aappling-usgs: Good Q. Actually, the one-to-many applications we've worked on in the last few months have used exactly this pattern, with one indicator file to represent many other files. Here's an example: https://github.com/USGS-R/lake-temperature-model-prep/blob/master/6_drivers/log/feather/6_fetch_new_cell_tasks.ind. This one is a complex example because it's an indicator file to represent indicator files for yml files, but a simpler example indicator file would look very similar, storing names and hashes of data files rather than other indicator files.

aappling-usgs:
I'm not sure what's happening with the build, given that the master branch is building without error and all you've done is added a vignette .Rmd file. Maybe the issue is occurring (proximately, anyway) because this vignette doesn't currently have a built version in the repo. You could try calling `devtools::build_vignettes()` locally, then manually moving the `doc` folder into `inst` (to make `inst/doc/.*`), then committing the new built-vignette files (.Rmd, .html, .R) for this vignette.

wdwatkins:
I got this error locally to actually: 
```

aappling-usgs:
Strange. Did you try the devtools::build_vignettes solution I suggested above?

wdwatkins:
Yes, just pushed those updates up.

aappling-usgs:
Hmm. I'm only seeing 3 total files pushed, was expecting 6:

- vignettes/target_types.Rmd
- inst/doc/target_types.Rmd
- inst/doc/target_types.R
- inst/doc/target_types.html
- inst/doc/task_plans.html
- docs/target_types.html

wdwatkins:
Looks like git was ignoring them since `doc` is ignored, and that must have made them still ignored in `inst/doc`

aappling-usgs:
Hey, it worked! Annoying, but Laura and I spend at least a day trying to work around it and had to give up, so I think this is the best bet for this and future vignettes for a while.

I wonder if our .gitignore is set up improperly. We do have `!inst/doc`, so I'd think things wouldn't be ignored in that folder, but maybe the `doc` line needs start/end line characters included, or maybe the order of `!inst/doc` and `doc` matters...something to look into someday.

aappling-usgs:
Have you made all the vignette changes you plan to make, @wdwatkins ?

wdwatkins:
Yes, ready for you to review again.  

I would guess the order of the two ignore lines matters.

wdwatkins:
I also wonder if there is some funniness going on in the task plans vignette that makes it not build the first time it runs, but then it uses existing temp files or something the second try, and it works.  There's no reason I can see for this PR to effect another vignette building or not.  I took a quick look and didn't see anything obvious though.  🤷‍♂

PR #118, opened 2019-08-29
which_dirty and why_dirty

aappling-usgs:
Resolves #57

Note that I downgraded .travis.yml to `warnings_are_errors: false` because I'm getting some warnings locally about package vignettes and think we've messed with those enough in the main branch that it's not worth dealing with them on this branch.

Issue #98, opened 2019-02-06
`scdel` should allow "deletion" of objects and group targets


aappling-usgs:
See also issue #15 (warnings on `scdel(object)`) and issues encountered in getting vignettes+tests to pass in PR #86 (e.g., https://travis-ci.org/USGS-R/scipiper/jobs/488817085#L1203)

aappling-usgs:
#15 was resolved sometime before today, and remake:::delete() seems to work fine for objects.

Group (fake) targets are never current and have no entries in the remake database, as far as I can tell. Therefore scdel just shouldn't bother attempting to `remake:::delete()` them. This is resolved by 19fc637

aappling-usgs:
Evidence that #120 fixes this issue:

example remake.yml:
```yml
targets:
  wqp_pull_string:
    command: c(I('hi mom!'))
  
  fake_target:
    depends:
      - wqp_pull_string
```

commands to build and delete the object and group=fake targets:
```r
> scmake('wqp_pull_string')
Starting build at 2019-08-30 17:16:32
<  MAKE > wqp_pull_string
[ BUILD ] wqp_pull_string                                  |  wqp_pull_string <- c("hi mom!")
Finished build at 2019-08-30 17:16:32
Build completed in 0.00 minutes
> scmake('fake_target')
Starting build at 2019-08-30 17:16:36
<  MAKE > fake_target
[    OK ] wqp_pull_string
[ ----- ] fake_target
Finished build at 2019-08-30 17:16:36
Build completed in 0.00 minutes
> scdel('wqp_pull_string')
(   DEL ) wqp_pull_string                                  |  rm("wqp_pull_string")
> scdel('fake_target')
> scmake('fake_target')
Starting build at 2019-08-30 17:16:45
<  MAKE > fake_target
[ BUILD ] wqp_pull_string                                  |  wqp_pull_string <- c("hi mom!")
[ ----- ] fake_target
Finished build at 2019-08-30 17:16:45
Build completed in 0.00 minutes
> scdel('fake_target')
> scdel('wqp_pull_string')
(   DEL ) wqp_pull_string                                  |  rm("wqp_pull_string")
> scdel('wqp_pull_string')
(       ) wqp_pull_string
```
(note that scdel of a fake target does nothing, which is pretty much consistent with remake's refusal to delete fake targets.)

PR #120, opened 2019-08-30
which_dirty and why_dirty

### LINKED ISSUE #57: which_dirty() and why_dirty() functions
Would be nice to know, before calling scmake(), which targets may get rebuilt and why. I think we can do this by drawing on the contents of build/status files, `scipiper::get_remake_status`, and maybe some enhancements to `scipiper::get_remake_status`.

aappling-usgs:
Resolves #57 and #98. See PR #118 for examples of the main functionality proposed in this PR, and #98 for the add-on for that bugfix (which I caught and fixed quickly while checking on the vignettes).

Now that `task_combiners` has been merged into `master` in the canonical repo, I'm going to close #118 (to merge these changes into `task_combiners`) in favor of this new PR to merge the changes into `master`.


limnoliver:
Just figured out what the `fixed` return is in `why_dirty` -- and that's an unexpected but welcomed feature! What exactly is it checking? Is it checking the fixed inputs of what you're checking, or fixed inputs of all dependencies?

limnoliver:
I'm testing these changes on the latest version of `lake-temperature-model-prep`. When I look at the output from `which_dirty()` I see: 

image

But then when I call `why_dirty()` on any of the outdated targets that are just R objects (e.g. `why_dirty('nldas_new_cells')` or `why_dirty('replete_time_range')`) it returns:

image

Does it have something to do with the fact that these are new (to me) targets? That seems likely since I can do `why_dirty('coop_parsers)` which is not new to me and it returns:

image

aappling-usgs:
> Just figured out what the `fixed` return is in `why_dirty` -- and that's an unexpected but welcomed feature! What exactly is it checking? Is it checking the fixed inputs of what you're checking, or fixed inputs of all dependencies?

Depends on the row - the first row is the hash of the combined fixed inputs of just the command for this target. Later rows where type='depends' are the fixed inputs for those targets.

aappling-usgs:
Good catch on the not-yet-built objects, thanks. I think I can fix that.

aappling-usgs:
I was able to reproduce the error with `nldas_new_cells` by building it once, then making a trivial change to the `cell_list_to_df` function, then asking why_dirty. I was able to fix it with a small change to `why_dirty()` that I'll push in a moment.
```r
> why_dirty('nldas_new_cells')
 Error in if (remake_object$targets[[row$name]]$type == "fake") { : 
  argument is of length zero 
> detach(package:scipiper, unload=TRUE)
> library(scipiper)
USGS Support Package: https://owi.usgs.gov/R/packages.html#support
> why_dirty('nldas_new_cells')
Since the last build of the target 'nldas_new_cells':
  * the function 'cell_list_to_df' used by the target has changed
# A tibble: 5 x 8
  type     name                         hash_old                         hash_new                         hash_mismatch dirty dirty_by_descent current
  <chr>    <chr>                        <chr>                            <chr>                            <lgl>         <lgl> <lgl>            <lgl>  
1 target   nldas_new_cells              0131d42786b22343eeceda7080192c07 0131d42786b22343eeceda7080192c07 FALSE         TRUE  FALSE            FALSE  
2 depends  nldas_cells                  1bf1a4c8fe8eeef3d631b5c487b980e7 1bf1a4c8fe8eeef3d631b5c487b980e7 FALSE         FALSE FALSE            TRUE   
3 depends  6_drivers/in/nldas_cells.rds 6ee456b5eae671353915601cba054c63 6ee456b5eae671353915601cba054c63 FALSE         FALSE FALSE            TRUE   
4 function cell_list_to_df              5c0aeba100e52d497e92aa3584ba6c79 5b2cdc9976712866b7267e9c96016672 TRUE          FALSE FALSE            TRUE   
5 function nldas_diff_cells             dbd5a4bfeb41accc1fac6da9729e27e5 dbd5a4bfeb41accc1fac6da9729e27e5 FALSE         FALSE FALSE            TRUE   
```

PR #121, opened 2019-09-05
Scbless

### LINKED ISSUE #61: option to assert that certain tasks are up to date
Sometimes a function or input changes and you just know that some targets that depend on it aren't going to change at all as a result. It'd be nice if you didn't have to actually rebuild those targets anyway just to persuade remake that they're up to date.

We might need separate implementations of this idea for task tables and for scmake more generally, or maybe we can just make a single change in scmake.

To implement, you could start by editing the hash of the changed dependency from the old value to the new value in the build/status file. Then call RDSify_build_status to push the yml info into remake's status format (rds in .remake folder). Maybe that's everything, or maybe there's a step I'm not thinking of right now?

aappling-usgs:
Resolves #61. See [test-scbless.R](https://github.com/USGS-R/scipiper/pull/121/files#diff-3e1fadc7ff6f996322f89e68bff69cdf) for examples. I went with making `scbless` and `sc_declare_current` both name the same function, because I think each may be more intuitive for a different set of people.


jread-usgs:
Nice work!

jzwart:
I tried out `scbless` on my current [pipeline ](https://github.com/jzwart/delaware-water-temp) for water temperature predictions in DRB, and it works great! 

I made `target_A` dirty, which would trigger a rebuild of `target_A` and `target_B`. I used `scbless('target_A')`, which made `target_A` current and did not trigger a rebuild of either `target_A` or `target_B` because `target_B` was no longer dirty by descent. 
If `target_A` was dirty and I used `scbless('target_B')`, this would still trigger a rebuild because `target_A` was still dirty and `target_B` was still dirty by descent as indicated by a very useful warning: 
```
these targets have been declared current but are still dirty by descent: target_B
```
👏 👏 💯 👏 👏

PR #126, opened 2020-01-14
housekeeping

aappling-usgs:
Delete one deprecated argument, rebuild with new roxygen, add index.html page for github.io


jread-usgs:
Looks good to me. Was your comment about my PR having 48 diff'd files today because you _only_ have 47 on this one? 🤣

aappling-usgs:
Hahaha! I was just only expecting 1 diffed file in your last PR based on our conversation =). Thanks for reviewing this one.

PR #127, opened 2020-02-03
v0.0.17

aappling-usgs:
Two small changes plus a version bump:

1) You can now call `scipiper:::get_dependency_status('targetname')` without writing all the code to get a remake_object. This is still an internal function but has been proving its worth for debugging recently.

2) In task plans, objects with complex names (e.g., 'nhdhr_23094-019867') were giving trouble in the combiner call, where remake failed to parse them. This PR adds tick quotes (\`) around objects in the combiner calls by default. You might want to skip this change for current built task plans, in which case you can specify `tickquote_combinee_objects = FALSE` in the call to `create_task_makefile` so that your task plan stays unchanged. If you do this, you'll get a warning about this argument being deprecated (already, even though this PR introduces it) because in the long run I think we'll be better off with always tickquoting.

### DIFF
--- a/R/create_task_makefile.R
+++ b/R/create_task_makefile.R
@@ -32,6 +32,7 @@
 #'   dummy target name instead, with suffix "_promise". This allows us to avoid cyclic 
 #'   dependencies. Naming convention for `_promise` variables is to drop any dir structure 
 #'   from the `final_targets`
+#' @param tickquote_combinee_objects TRUE by default
 #' @return the file name of the makefile that was created, 
 #'   or the string output (if makefile = NULL)
 #' @export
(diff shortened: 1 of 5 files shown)


R/create_task_makefile.R
```diff
@@ -130,7 +136,7 @@ create_task_makefile <- function(
     if (('remake' %:::% 'target_is_file')(target, file_extensions = pos_extensions)){
       paste0("'", target, "'")
     } else {
-      target
+      if(tickquote_combinee_objects) paste0("`", target, "`") else target
```
jread-usgs: well that was pretty easy/clean

aappling-usgs: yeah, i should have looked before worrying/complaining =)

jread-usgs: did you want `tickquote_combinee_objects` or `tickquote_combine[r]_objects`?

aappling-usgs: i intended `combinee` because it's referring to objects that get combined. What's your thinking?

jread-usgs: haha, sounds good. I thought maybe it was a typo, but I get it now.

aappling-usgs:
@jread-usgs Since you've already looked, do you mind reviewing/approving/merging this PR?

Issue #116, opened 2019-08-26
task combiners: allow custom inputs to the `finalize_funs`


lindsayplatt:
Use-case for arguments being passed to the combiner functions other than the list of files to combine.

The `vizstorm-GIF` use-case is not perfect for task combiners since it needs to combine resulting frames from multiple makefiles instead of combining all the resulting frames from one makefile. However, if we had not added `intro` and `outro` frame makefiles, then I would have been able to use task combiners and my `finalize_funs` would have been `combine_animation_frames`, which is the function that creates the GIF. It is important for this function to be able to accept additional arguments (such as framestep) that can be parameterized using `viz_config`. This is currently not possible with how `finalize_funs` is implemented.

PR #141, opened 2020-07-13
add combine_to_tibble

### LINKED ISSUE #140: create another package-level combiner: "combine_to_tibble" or otherwise returns an object
`combine_to_ind(ind_file, ...)` returns a file. But there are similar cases where you want to do the same thing but return an R object (e.g., a tibble with `filepath` and `hash` columns). 

perhaps something like `combine_to_tibble(...)` (or `_to_table` to avoid the tibble/data.frame naming)?

I thought maybe this would be what `coming_to_ind` could do if you specify `ind_file = NULL`, but that wouldn't work with the current version of combiner functions, which don't have much flexibility for arguments.

aappling-usgs:
Resolves #140 with this behavior:

```
> tfiles <- tempfile(pattern=as.character(1:3))
> file.create(tfiles)
[1] TRUE TRUE TRUE

> combine_to_tibble()
Error in length(...) : 0 arguments passed to 'length' which requires 1

> combine_to_tibble(tfiles[1])
# A tibble: 1 x 2
  filename                                                       hash                      
  <chr>                                                          <chr>                     
1 /var/folders/_0/fbg0ffkj6z3fb_7jvm8wg86r002c5h/T//RtmpjELF9G/… d41d8cd98f00b204e9800998e…

> combine_to_tibble(tfiles[1:3])
# A tibble: 3 x 2
  filename                                                       hash                      
  <chr>                                                          <chr>                     
1 /var/folders/_0/fbg0ffkj6z3fb_7jvm8wg86r002c5h/T//RtmpjELF9G/… d41d8cd98f00b204e9800998e…
2 /var/folders/_0/fbg0ffkj6z3fb_7jvm8wg86r002c5h/T//RtmpjELF9G/… d41d8cd98f00b204e9800998e…
3 /var/folders/_0/fbg0ffkj6z3fb_7jvm8wg86r002c5h/T//RtmpjELF9G/… d41d8cd98f00b204e9800998e…
```


R/scmake.R
```diff
@@ -351,6 +351,33 @@ combine_to_ind <- function(ind_file, ...){
...
+#' 
+#' @details light wrapper on `sc_indicate`
+#' @export
+#' @examples 
+#' tfiles <- tempfile(pattern=as.character(1:3))
+#' file.create(tfiles)
+#' combine_to_tibble(tfiles[1])
+#' combine_to_tibble(tfiles[1:2])
+#' combine_to_tibble(tfiles[1:3])
+combine_to_tibble <- function(...){
```
jread-usgs, line 366: Interesting and more comprehensive solution than what I did quickly [here](https://github.com/collnell/gage_age_gif/pull/1/files#diff-a3f350b879f5de0415f9e1c1655b6cabR18).

I see good reason to use `sc_indicate` for all file hashing activities, but the file length differences does require some odd work arounds. Good solution though :+1:

This looks good but does bring up the question on whether we want a few more tests in this package as a future goal? In this case, I would see verifying that `combine_to_tibble(tfiles[1]) == combine_to_tibble(tfiles)[1,]` is TRUE

aappling-usgs: Hah, good point that we could just bypass `sc_indicate`...I think I'll stick with it but agree it's clunkier in some ways.

I'll add a test.

aappling-usgs:
Ugh, nothing is ever simple. Should this function support combining of objects? I guess I'm going to leave it out for now, because I got this far in rewriting it before realizing I couldn't rely on `remake:::target_is_file` or `fetch` without also allowing custom arguments in combiners (#116 ):
```r
    tibble(name=c(...)) %>%
      mutate(hash = purrr::map(name, function(nm) {
        if(remake:::target_is_file(nm)) { # but what if it's a non-default file extension?
          unname(tools::md5sum(nm))
        } else {
          digest::digest(fetch(nm)) # and what if it's a non-default remake file?
        }
      }))
```

aappling-usgs:
OK @jread-usgs , I added some tests and did a bit of cleanup for edge cases. See also comment above, but if the current changes seem like they can stand alone, let's just merge and be done with it for now.

tests/testthat/test-combiners.R
```diff
@@ -0,0 +1,12 @@
+context("Combine task targets into summary files or objects")
+
+test_that("combine_to_tibble works on all reasonable numbers of files", {
+  tfiles <- tempfile(pattern=as.character(1:3))
+  for(i in 1:3) readr::write_lines(i, path=tfiles[i])
```
jread-usgs, line 5: was hoping you'd catch that the files would need to be different for this test :)

PR #148, opened 2020-11-20
update readr::write_lines fxn according pkg update in v1.4.0

lindsayplatt:
I started getting this warning when I updated `readr`. This should fix that from happening.

```r
Warning message:
The `path` argument of `write_lines()` is deprecated as of readr 1.4.0.
Please use the `file` argument instead.
```


aappling-usgs (approved):
Thanks!

aappling-usgs:
Oh - could you slip a version increment in there, @lindsayplatt ?

PR #154, opened 2021-12-29
Specify path for S3 objects when checking bucket

wdwatkins:
When buckets have lots of objects in them, they might not make the initial 1000 objects returned by `get_bucket`


jread-usgs:
Does this need a reviewer assigned?

wdwatkins:
@lindsayplatt Can you take a quick look maybe?  This is a very small but sometimes important change.

lindsayplatt:
Just to make sure I understand: there is a 1000 object limit on what is returned from `get_bucket()`. We are assuming that buckets with >1000 objects are organized into subdirs and adding in the `prefix` argument, so that we are just using `get_bucket()` for objects _within_ the subdir of the data we are looking for. If a subdir had >1000 objects, this still would cause an issue, right?

Do you have an example S3 bucket that has this issue?

wdwatkins:
Yep, that covers it.  In the event of > 1000 objects in a subdir we could just add `max = Inf` as well.  

I think I encountered this with the bucket we use for analytics log data.  It is mostly used for storing logs so has a couple thousand files in it, but it was also convenient to use it for a few cached files that are in an old scipiper workflow for building a quarterly report.  Probably unlikely you would encounter this when a bucket is 1:1 with a scipiper workflow.

R/s3.R
```diff
@@ -104,7 +104,8 @@ s3_put <- function(
   aws.signature::use_credentials(profile = s3_config$profile)
   
   # determine whether and where the remote file exists
-  bucket_contents <- aws.s3::get_bucket_df(bucket = s3_config$bucket)
+  bucket_contents <- aws.s3::get_bucket_df(bucket = s3_config$bucket,
+                                           prefix = dirname(data_file))
```
lindsayplatt, line 108: My only comment would be to add either comments, warnings, or that `max=Inf` now in order to save us future headaches if this pops up again.

One idea: if the bucket contents that comes back hits that 1000 item limit, we could send the user a warning to check that all their contents made it.
```suggestion
                                           prefix = dirname(data_file))
if(length(bucket_contents) >= 1000) warning('The S3 bucket returned items around the max limit (1000). Please verify that all your expected items were returned')
```

wdwatkins: Good idea, I will add something similar

wdwatkins: added max = Inf here

PR #157, opened 2022-07-13
bump version increment

wdwatkins:
For previous PR so there is a version to reference


wdwatkins:
self merging since this is not a code change

PR #158, opened 2022-10-25
Allow use of EC2 metadata for S3 credentials

jesse-ross:
## Motivation
I'm adapting Vizlab's [gage-conditions-gif](https://github.com/usgs-vizlab/gage-conditions-gif) pipeline to run in an ECS context, so that we can automate the generation of the flow-tiles visualization on a monthly basis. Currently, the code assumes that credentials are stored locally in a `~/.aws/credentials` file.

## Summary of Changes
This adds an option to skip looking for credentials on the local filesystem, and instead pull them from EC2 metadata.
1. The `aws.ec2metadata` package is now suggested for `scipiper` installation, and it is required for running the `s3` functions.
2. The `s3_get()` and `s3_put()` functions now have an argument `use_local_aws_credentials`. If this argument is `FALSE`, then EC2 credentials are used. The default value is `TRUE`, so existing pipelines which use the `s3` functions can run unchanged.

## Validation of Changes
[My fork of gage-conditions-gif](https://github.com/jesse-ross/gage-conditions-gif/tree/awsify) uses this version of `scipiper` (installation is [here](https://github.com/jesse-ross/gage-conditions-gif/blob/awsify/docker/Dockerfile#L44-L46)). It runs correctly on ECS, where I run it with a `.Rprofile` which sets `options(scipiper.use_local_aws_credentials = FALSE)` (here are the [CloudWatch logs](https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logsV2:log-groups/log-group/$252Fecs$252Fgage-conditions-task/log-events/ecs$252Fbuild-stats$252Feab6ed3e14454c918918cbaa4019e548), log in as developer to see). It also runs locally when that `.Rprofile` is not in place or sets `options(scipiper.use_local_aws_credentials = TRUE)`

## Outline of what is/is not expected from the reviewer
I don't have much `scipiper` experience, so any input at all would be welcome!

### DIFF
--- a/R/s3.R
+++ b/R/s3.R
@@ -72,7 +72,7 @@ s3_config <- function(bucket, profile='default', config_file=getOption("scipiper
 #'   project-specific configuration information for S3
 #' @param ind_ext the indicator file extension to expect at the end of
 #'   remote_ind
-#' @param use_local_aws_credentials logical If TRUE, calls to this function will rely on credentials in `~/.aws/credentials`; if FALSE, credentials will be gotten from EC2 metadata. Using credentials from EC2 metadata is appropriate when running on EC2 or ECS.
+#' @param use_local_aws_credentials If TRUE (the default), then calls to this function will rely on credentials in `~/.aws/credentials`; if FALSE, credentials will be gotten from EC2 metadata. Using credentials from EC2 metadata is appropriate when running on EC2 or ECS.
 #'   remote_ind
 #' @export
 s3_put <- function(
@@ -81,7 +81,7 @@ s3_put <- function(
   dry_put=getOption("scipiper.dry_put"),
   config_file=getOption("scipiper.s3_config_file"),
   ind_ext=getOption("scipiper.ind_ext"),
-  use_local_aws_credentials = TRUE) {
+  use_local_aws_credentials = getOption("scipiper.use_local_aws_credentials")) {
   
   # check arguments
   mock_get <- match.arg(mock_get)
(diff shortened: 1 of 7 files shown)


jesse-ross:
I've asked @wdwatkins for review here, since I think he has spent the most time on these `s3` functions, but the suggested reviewers are @aappling-usgs or @jordansread, and I'd also be really interested if either of you anything to say!

jordansread (commented):
Looks good to me! Just adding a couple of light comments. David is more familiar with this function than I am. 

A couple of suggestions:
 - can you increment the minor version number of the `scipiper` pkg? Makes sense to me to do this since you are adding a new pkg in this case and modifying the s3_put function args. 
 
 - Too picky on this given this is a legacy pkg, but in a future similar situation, something like this may be a touch more aligned with the other patterns in scipiper for default args in these put/get functions:
```r
..., use_local_aws_credentials = getOption("scipiper.use_local_aws_credentials") 
```
so that someone could set the pkg-level arg w/o changing their function calls. But again, too picky in this case given the legacy status of the pkg.

jesse-ross:
@jordansread I implemented your suggestion for setting a default option; thanks again!

@wdwatkins I'm removing the Draft marker on this. No big rush on reviewing it.

R/s3.R
```diff
@@ -177,7 +184,9 @@ s3_get <- function(ind_file, verbose = FALSE,
   if(verbose) {message("Downloading ", data_file, " from S3")}
   
   s3_config <- yaml::yaml.load_file(config_file)
-  aws.signature::use_credentials(profile = s3_config$profile)
+  if (use_local_aws_credentials) {
+    aws.signature::use_credentials(profile = s3_config$profile)
```
wdwatkins, line 188: What happens if you don't add these?  I think this is run already when `aws.signature` is loaded, and grabs the default profile from the default location if no environment vars are set, per the help file:

> use_credentials uses credentials from a profile stored in a credentials file to set the environment variables used by this package. It is called by default during package load if the AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY environment variables are not set.

But I don't think this would hurt anything to make sure the expected profile is being used, if you wanted to use something besides the default AWS profile and didn't already set the environment variable.

wdwatkins (commented):
This look good to me, just the one comment on if the explicit `use_credentials` is necessary.

R/s3.R
jesse-ross: If I don't add the `if` logic here, the s3 operations fail on ECS with something like "could not find /root/.aws/credentials".

I am a little bit hazy on exactly what is happening under the hood, but it seems that `aws.ec2metadata` may be providing `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`, so that `aws.signature::read_credentials()` doesn't need to be called? Whereas if we call that function, it obscures the information from `aws.ec2metadata`?

In the ECS case, I did have to work around not having my region in a profile, which I did by [just setting it as an environment variable](https://github.com/jesse-ross/gage-conditions-gif/blob/awsify/aws/run-pipeline.sh#L3). Apart from region and credentials, do you know other important things which would go into the profile? I.e. do you think I need to be figure out a way to read the profile but not clobber the credentials when doing so?

wdwatkins: Right, yeah `use_credentials` won't work on EC2.  I guess what I'm asking is, is `use_credentials` still needed on a local machine, since it sounds like it runs when the aws.signature package is loaded already?  

I don't think there is other profile information you would need besides region.  I think its reasonable to expect a user to mostly handle credentials outside of this package if needed.

jesse-ross: Got it - yeah, it looks like if you're using a default profile, then there'd be no need for the `use_credentials()` call. But it's still nice to have this code for the case where the user wants to specify a non-default profile in the `s3_config.yml`.

